PR #17, opened 2019-01-28
Documentation: Added derivatives structure to BIDS

alexfoias:
Added derivatives structure


alexfoias:
I was thinking to stick to the a similar field separation as BIDS. `_` is used as separator between different fields and `-` is used as separator within a field.

@perone what do you think is better from the API point of view ?

alexfoias:
@jcohenadad The `duke/sct_testing/bids/test_dataset/bids_unf_sct_ver1` dataset already uses `-` as in-field separator. For the moment, the API parses the full content of derivatives of a specific modality for an unique subject.

jcohenadad:
@alexfoias ok let's go with `seg-manual` then :-)

perone:
For the derivatives, I'm just parsing the prefix to find the related derivatives (because they have a 1:1 relationship with the volumes outside the derivatives folder) but we don't require any format after that, so as long as we use something that is the same for the same content it will be fine.

PR #18, opened 2019-02-08
Created new wrapper to process BIDS architecture

Nicolas-Pinon:
Special wrapper compatible with BIDS structure as described [here](https://github.com/sct-pipeline/spine_generic/tree/alexfoias/derivatives_structure)
Comes with its own parameter file


Nicolas-Pinon:
I thought that in the BIDS structure "derivatives" really meant "files that were derived from the originals", thus I did not include the untouched gre-t1w

jcohenadad:
> I thought that in the BIDS structure "derivatives" really meant "files that were derived from the originals", thus I did not include the untouched gre-t1w

True, but to be pragmatic, what's going to happen with this project is: we develop something that works, we publish, people will want that we share the data, so we will make a "clean" package that only contains the data used for this study (i.e., all volumes + the seg). If we share the entirety of the dataset it will become heavy (several GB) and messy (same files located at two different places-- one processed, one not processed). The community interested in this dataset doesn't care about the unprocessed data-- all they want is the data used to generate/test the deep learning model.

Nicolas-Pinon:
Then do we still stick with the BIDS data format as output ? I feel like the BIDS format is kinda "made" to have all the datas in it, but I might be wrong.
I completely see your point though

generate_ground_truth.sh
```diff
@@ -41,28 +41,39 @@ if [ -e "${file_t1w}_seg_manual.nii.gz" ]; then
...
-rm ${ofolder}/*warp* #delete warping fields
-rm ${ofolder}/"${file_t1w_mts}_reg.nii.gz" #delete the t1w MTS registered to the last image (here T1w)
+rm ${ofolder_reg}/*warp* #delete warping fields
+
+# copying the T1w_mts file, which everything else is registered into :
+rsync -avzh "${file_t1w_mts}.nii.gz" ${ofolder_reg}/
+
+# copying the json files :
+rsync -avzh "${file_t1w_mts}.json" ${ofolder_reg}/
+rsync -avzh "${file_mton}.json" ${ofolder_reg}/
```
jcohenadad: BIDS requires that the json files have the exact same name (everything before the extension) as the NIFTI file. Here, you will be missing the "_reg"

Nicolas-Pinon:
> @Nicolas-Pinon why did you create a new `parameters_bids.sh` and `run_process_BIDS.sh`?

Because at the start we created this script in the same repo (this repo) as the spine_generic pipeline. I created a new parameter file not to confuse it with the old one, and for the wrapper, I had to create a new one, otherwise the outputed dataset was not BIDS compatible (see first issue we run onto with Christian)

generate_ground_truth.sh
```diff
@@ -11,30 +11,24 @@
...
 # Create BIDS architecture
-# mkdir derivatives
-# mkdir derivatives/labels
-# mkdir derivatives/labels/${sub}
-# mkdir derivatives/labels/${sub}/anat
-ofolder="derivatives/labels/${sub}/anat"
+mkdir -p ${PATH_OUTPUT_wSITE}/${sub}/anat
+mkdir -p ${PATH_OUTPUT_wSITE}/derivatives/labels/${sub}/anat
+ofolder_seg="${PATH_OUTPUT_wSITE}/derivatives/labels/${sub}/anat"
+ofolder_reg="${PATH_OUTPUT_wSITE}/${sub}/anat"
```
jcohenadad: @Nicolas-Pinon why do you need this variable if everything is created locally?

jcohenadad: nevermind-- i just realized that you cd to the input folder (not to the output folder)

generate_ground_truth.sh
Nicolas-Pinon: Done

jcohenadad:
Scripts moved here: https://github.com/neuropoly/ivado-medical-imaging/tree/master/prepare_data

PR #31, opened 2019-05-02
added automatic parsing of json for TR and FA

### LINKED ISSUE #27: MTsat calculation: use JSON values instead of hard-coded ones
In [this line of code](https://github.com/sct-pipeline/spine-generic/blob/master/processing/process_data.sh#L167), TR and flip angle are currently hard-coded, but ultimately those values should be extracted from the json file. 

This could be done via bash or a python reader. Either way is fine.

Nicolas-Pinon:
Didn't have time to test the whole script yet, not really elegant but should do the trick

Fixes #27

### DIFF
--- a/processing/process_data.sh
+++ b/processing/process_data.sh
@@ -85,6 +85,12 @@ segment_gm_if_does_not_exist(){
   fi
 }
 
+get_field_from_json(){
+  local file="$1"
+  local field="$2"
+  echo `grep $field $file | sed 's/[^0-9]*//g'`
+}
+
 
 # SCRIPT STARTS HERE
 # ==============================================================================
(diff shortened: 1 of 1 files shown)


jcohenadad:
thanks for the quick implementation 👍 i will test asap

jcohenadad:
@Nicolas-Pinon your code doesn't work-- issue if more than 2 numbers-- could you pls find a workaround? thx

Nicolas-Pinon:
Fixed the parsing, should work now (+ solution more elegant), doing the testing now on abbey

jcohenadad:
> Fixed the parsing, should work now (+ solution more elegant), doing the testing now on abbey

Awesome 👍 Let me know when you finish the testing so we can merge. Thanks!

Nicolas-Pinon:
Everything ran just fine on my end ! :)

jcohenadad:
Awesome! I'll clean up a bit (make function to avoid duplication) and merge. Thanks!

PR #40, opened 2019-05-13
Compatibility fix and minor improvements on the figure generation

jcohenadad:
- Refactored to make figure generation compatible with the new BIDS structure, where all sites and subjects are organized in a flat hierarchy
- possibility to display individual subject plot


charleygros (approved):
Tested!
Sounds good to me
Great cleaning in `generate_figure`

PR #19, opened 2019-02-18
BIDS dataset validator

alexfoias:
This validator is a modified version of [bids-validator](https://github.com/bids-standard/bids-validator).

-  **dataset_validator** is used to check an individual BIDS dataset.
   Usage: `python dataset_validator.py -d PATH_TO_INDIVIDUAL_BIDS_DATASET`

- **wrapper_dataset_validator** is used to loop within a root path which hosts multiple BIDS datasets. 
   Usage: `python wrapper_dataset_validator.py -d PATH_TO_ROOT_BIDS_DATASETS`

PR #48, opened 2019-05-22
WIP acquisition parameters validator

alexfoias:
The aim of this PR is to provide a validator for the acquisition parameters.

@jcohenadad Do we have to define different value for various vendors for the same modality ? How harsh do we have to be with the actual values compared to the protocol ones (ex. EchoTime = 0.003702 instead of 0.00372)?


jcohenadad:
duplicate of https://github.com/spine-generic/data-single-subject/pull/3?

as mentioned in #19 we should centralize those validator into a single repository (i suggest here), and import the package during CI

alexfoias:
This implementation is obsolete, another PR will be opened.

PR #49, opened 2019-05-28
create_mosaic: impose isotropic resolution

### LINKED ISSUE #45: create_mosaic: Fix problem of AP scaling
which could be due to anisotropic voxel resolution. See: https://github.com/sct-pipeline/spine-generic/pull/41#issuecomment-493053849

charleygros:
We observed some scaling problem in the AP direction of sagittal images, see #45.

This PR is fixing this issue by resampling anisotropic sagittal images with isotropic resolution.

Note: Please do not review this PR yet. I need to properly test it first.

Fixes #45.


Nicolas-Pinon (approved):
Tested, runs properly, seems to do the trick, but not tested on many subjects tho

Issue #47, opened 2019-05-21
generate_figure: Use friendlier names for sites


jcohenadad:
notably, change these:
- GE scanner with current name “brno” to “UHBrno" (University Hospital Brno) or shorter version “UHB" 
- Siemens scanner with current name “brnoPrisma” to “CEITECBrno" (Central European Institute of Technology Brno) or shorter version “CEITEC"

valosekj:
I proposed solution directly inside`generate_figure.py` script. 

But most correct fix of this issue would be to rename source files (nii and json files) in dataset at openneuro.org website, rerun `process_data.sh` and then change names of sites in: https://github.com/sct-pipeline/spine-generic/blob/d51799c7339c1b066d098a1427d9aed7bba3e177/processing/generate_figure.py#L91-L92

We can also consider to rename some another sites which have name of the scanner inside site name (like juntendo750w, vuiisAchieva, beijingPrisma,...) because scanner name is included inside bar plots.

valosekj:
@jcohenadad for final paper´s figures, we could invigorate discussion about names for sites. The easiest solution is probably using "conversion dict", something how I outlined here https://github.com/sct-pipeline/spine-generic/pull/55/files

PR #55, opened 2019-11-09
fix site names (issues #47)

valosekj:
Proposed solution for #47

### DIFF
--- a/processing/generate_figure.py
+++ b/processing/generate_figure.py
@@ -514,7 +514,13 @@ def main():
         ax = label_bar_model(ax, bar_plot, model_sorted)  # add ManufacturersModelName embedded in each bar
         plt.setp(ax.xaxis.get_majorticklabels(), rotation=45, ha="right")  # rotate xticklabels at 45deg, align at end
         plt.xlim([-1, len(site_sorted)])
-        ax.set_xticklabels([s for s in site_sorted])  # add space after the site name to allow space for flag
+
+        # Change site names
+        site_sorted_fix = site_sorted.copy()
+        site_sorted_fix[site_sorted == 'brno'] = 'brnoUH'
+        site_sorted_fix[site_sorted == 'brnoPrisma'] = 'brnoCEITEC'
+
+        ax.set_xticklabels([s for s in site_sorted_fix])  # add space after the site name to allow space for flag
         # ax.get_xaxis().set_visible(True)
         ax.tick_params(labelsize=15)
         # plt.ylim(ylim[contrast])


alexfoias:
do we only need to change the following ?

- brno → brnoUH
- brnoPrisma → brnoCEITEC

jcohenadad:
> do we only need to change the following ?
> 
> * brno → brnoUH
> * brnoPrisma → brnoCEITEC

yes i think so

alexfoias:
Okay. Are we only changing the info in the column `institution_id` or do we need to rename the subjects as well ?

jcohenadad:
> Okay. Are we only changing the info in the column `institution_id` or do we need to rename the subjects as well ?

let's go with the clean approach: change subject id as well

alexfoias:
I rechecked the names that I suggested on top with regard to the ones [here](https://github.com/sct-pipeline/spine-generic/issues/47#issue-446711620). 
The correct naming convention should be this one:
- brno → UHBrno
- brnoPrisma → CEITECBrno

@jcohenadad Could you please confirm before I go ahead in propagating the changes ? thanks.

jcohenadad:
> I rechecked the names that I suggested on top with regard to the ones [here](https://github.com/sct-pipeline/spine-generic/issues/47#issue-446711620).
> The correct naming convention should be this one:
> 
> * brno → UHBrno
> * brnoPrisma → CEITECBrno
> 
> @jcohenadad Could you please confirm before I go ahead in propagating the changes ? thanks.

@alexfoias what convention are you referring to?  i'm looking at the list of sites [here](https://openneuro.org/datasets/ds001919/versions/1.0.5), and it seems that the camelCase (with first letter small) is the adopted convention. Moreover, cases where a city has more than one center (e.g. oxford), the city is first, then the center. Which means that we should probably have something like this: `brnoUhb` and `brnoCeitec`. @valosekj are you OK with this?

valosekj:
@jcohenadad yes, it looks fine for me. Maybe, better than `brnoUhb` could be `brnoUh` (without b at the end). 'brnoUh' stands for University Hospital Brno, so it is not probably necessary to include the last b (meaning first letter from Brno city) because Brno (=city) is already included in the name. I only quickly ask colleagues from there for confirmation. I will let know.
`brnoCeitec` is fine, it means Central European Institute of Technology (CEITEC) Brno.

jcohenadad:
> @jcohenadad yes, it looks fine for me. Maybe, better than `brnoUhb` could be `brnoUh` (without b at the end). 'brnoUh' stands for University Hospital Brno, so it is not probably necessary to include the last b (meaning first letter from Brno city) because Brno (=city) is already included in the name. I only quickly ask colleagues from there for confirmation. I will let know.
> `brnoCeitec` is fine, it means Central European Institute of Technology (CEITEC) Brno.

yes i got it about the duplicated b, but sometimes institution names are immutable (e.g. we write "university of oxford, oxford, uk" not "university of, oxford, uk").

valosekj:
@jcohenadad So, after discussion with colleagues, `brnoUhb` and `brnoCeitec` is fine for us. Thank you 👍

alexfoias:
We have changed the names in 1.0.7. We can close this PR.

jcohenadad:
change implemented in this release: https://github.com/sct-pipeline/spine-generic/releases/tag/v1.1

PR #60, opened 2020-05-24
Scaling adjustment, bisection line added, grid moved to background

valosekj:
Included changes:
- added dynamic axis scaling of individual subplots
- added bisection (diagonal) line
- grid moved to background

I also increased size of linear fit equation, it resulted in conflict with vendor label, so I moved vendor label to lower-right corner.

fig_t1_t2_agreement_per_vendor

Fixes: https://github.com/sct-pipeline/spine-generic/issues/57

### DIFF
--- a/processing/generate_figure.py
+++ b/processing/generate_figure.py
@@ -618,9 +618,12 @@ def main():
                     label=vendor)
         # Define vendor name position
         legend = ax.legend(loc='lower right', handletextpad=0, fontsize=FONTSIZE)
+        # Change box's frame color to black (to be same as box around linear fit equation)
+        frame = legend.get_frame()
+        frame.set_edgecolor('black')
         ax.add_artist(legend)
         # Dynamic scaling of individual subplots based on data
-        offset = 6
+        offset = 8
         lim_min = min(min(x), min(y))
         lim_max = max(max(x), max(y))
         plt.xlim(lim_min - offset, lim_max + offset)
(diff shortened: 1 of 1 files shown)


valosekj:
Could be done:

fig_t1_t2_agreement_per_vendor

jcohenadad:
cool! but could you make the grid "squary"? (currently it's a rectangle)

also, can you try putting a box around the y=ax+b? the grid is somewhat distracting

valosekj:
Sure! I added box around linear fit equation and changed vendor box's frame color to be same as former. Also, I used `plt.gca().set_aspect("equal")` to enforce grid to be squared.
I also increased space between title and subplots.

fig_t1_t2_agreement_per_vendor

processing/generate_figure.py
```diff
@@ -603,28 +604,57 @@ def main():
...
-        plt.ylim(45, 100)
-        plt.gca().set_aspect('equal', adjustable='box')
+        # Define vendor name position
+        legend = ax.legend(loc='lower right', handletextpad=0, fontsize=FONTSIZE)
+        # Change box's frame color to black (to be same as box around linear fit equation)
+        frame = legend.get_frame()
+        frame.set_edgecolor('black')
+        ax.add_artist(legend)
+        # Dynamic scaling of individual subplots based on data
+        offset = 8
```
valosekj, line 626: By changing this, it is possible to "zoom" more to data. But I noticed that not for all numbers is the grid "squared".

jcohenadad:
Almost there! I would decrease the font size of the equation (e.g. use the same as the x/y-ticklabel).

Also, can you try to put the equation + R2 in red? just to see if it is more intuitive to interpret the red line (because it is linked with the equation).

Also, i would bring the three plots closer (there is too much unnecessary space between them).

valosekj:
Decreased spacing between subplots. Smaller equation font and red color. For me, it looks reasonable to have it in red color.

fig_t1_t2_agreement_per_vendor

Issue #63, opened 2020-05-27
generate_figure.py input parameters


valosekj:
The [generate_figure.py](https://github.com/sct-pipeline/spine-generic/blob/master/processing/generate_figure.py) script usage could be more convenient if results directory could be set as input parameter, how is proposed in this TODO:

https://github.com/sct-pipeline/spine-generic/blob/aa697cd96f664c593d7f69670605c43f4bc9fe2f/processing/generate_figure.py#L450

Maybe, also data directory which contains participants.tsv file could be set as input parameter:

https://github.com/sct-pipeline/spine-generic/blob/aa697cd96f664c593d7f69670605c43f4bc9fe2f/processing/generate_figure.py#L214

I can do it, if it looks okay for you @jcohenadad.

jcohenadad:
> The generate_figure.py script usage could be more convenient if results directory could be set as input parameter.

yup, but i would keep default behaviour as it is--  more convenient (i.e. no need to type additional stuff in the CLI)

> Maybe, also data directory which contains participants.tsv file could be set as input parameter:

this is somewhat less needed, thanks to the recent change that copies this file under results/data: https://github.com/sct-pipeline/spine-generic/commit/96a772c6546233d8271dd441b5f78d3f9f8956f4

valosekj:
> > The generate_figure.py script usage could be more convenient if results directory could be set as input parameter.
> 
> yup, but i would keep default behaviour as it is-- more convenient (i.e. no need to type additional stuff in the CLI)
> 

I understand. So, it could be an optional argument for situation when user would like to enforce "another results" directory.

PR #64, opened 2020-05-28
generate_figure.py: make "results" an optional input param; other improvements

### LINKED ISSUE #111: Update doc with new behaviour without paramers.sh
here: https://spine-generic.readthedocs.io/en/latest/documentation.html#generate-figures

the doc instructs to use a parameters.sh file, but we don't have that anymore. So, doc should be updated

valosekj:
This PR introduces the flag `-path-results` to the csv folder, for more convenient use. Fixes: #63 

Additional improvements include:
- Clarification of computation for intra/inter-COV
- Update documentation to reflect the drop of the parameters.sh file. Fixes #111


processing/generate_figure.py
```diff
@@ -199,6 +199,11 @@ def get_parameters():
         required=False,
         help="Display the value of each individual subject as a red dot.",
         default=1)
+    parser.add_argument(
+        '-results',
```
jcohenadad, line 203: i would call it "-path-results"  to be explicit

jcohenadad:
@valosekj is this ready to be reviewed?

valosekj:
> @valosekj is this ready to be reviewed?

@jcohenadad Yes, it is. I have tested it again just now and it works fine.

valosekj:
> @valosekj as you can see i removed the duplicated usage description between the function header and the usage-- hope that's ok with you

Yes, it is ok. Thank you 👍

PR #87, opened 2020-06-23
Added sh script for manual correction

### LINKED ISSUE #85: Add manual correction scripts to the repository
instead of asking people to copy/paste the content of the scripts [here](https://spine-generic.readthedocs.io/en/latest/documentation.html#segmentation), let's just add those scripts to this repository, and ask users to run them.

valosekj:
fixes #85

### DIFF
--- a/docs/documentation.rst
+++ b/docs/documentation.rst
@@ -298,52 +298,8 @@ If you spot segmentation issues, manually fix them using the procedure described
 below. Also see the video tutorial after the procedure.
 
 - Go to the ``results/data`` folder
-- Create the file ``fix_seg.sh`` and copy/past the code below:
-
-.. code-block:: bash
-
-  #!/bin/bash
-  # Folder to output the manual labels
-  PATH_SEGMANUAL="../../seg_manual"
-  mkdir $PATH_SEGMANUAL
-  # List of files to correct segmentation on
-  FILES=(
-  sub-amu02_acq-T1w_MTS.nii.gz
-  sub-beijingGE04_T2w_RPI_r.nii.gz
-  sub-brnoPrisma01_T2star_rms.nii.gz
-  sub-geneva04_dwi_crop_moco_dwi_mean.nii.gz
-  )
-  # Loop across files
-  for file in ${FILES[@]}; do
-    # extract subject using first delimiter '_'
-    subject=${file%%_*}
-    # check if file is under dwi/ or anat/ folder and get fname_data
-    if [[ $file == *"dwi"* ]]; then
-      fname_data=$subject/dwi/$file
-    else
-      fname_data=$subject/anat/$file
-    fi
-    # get fname_seg depending if it is cord or GM seg
-    if [[ $file == *"T2star"* ]]; then
-      fname_seg=${fname_data%%".nii.gz"*}_gmseg.nii.gz${fname_data##*".nii.gz"}
-      fname_seg_dest=${PATH_SEGMANUAL}/${file%%".nii.gz"*}_gmseg-manual.nii.gz${file##*".nii.gz"}
-    else
-      fname_seg=${fname_data%%".nii.gz"*}_seg.nii.gz${fname_data##*".nii.gz"}
-      fname_seg_dest=${PATH_SEGMANUAL}/${file%%".nii.gz"*}_seg-manual.nii.gz${file##*".nii.gz"}
-    fi
-    # Copy file to PATH_SEGMANUAL
-    cp $fname_seg $fname_seg_dest
-    # Launch FSLeyes
-    echo "In FSLeyes, click on 'Edit mode', correct the segmentation, then save it with the same name (overwrite)."
-    fsleyes -yh $fname_data $fname_seg_dest -cm red
-  done
-
-- Make this script executable:
-
-  .. code-block:: bash
-
-     chmod 775 fix_seg.sh
-
+- Edit list of files to correct segmentation on, i.e. edit ``FILES`` variable in ``manual_correction.sh`` script
+- Run the ``manual_correction.sh`` script
 - In the QC report, enter the string "deepseg" to only display segmentation results.
 - Review all segmentations. Use the keyboard shortcuts up/down arrow to switch between
   subjects and the left arrow to toggle overlay.
--- a/processing/manual_correction.sh
+++ b/processing/manual_correction.sh
@@ -2,29 +2,70 @@
 
 # Run this script in results/data folder
 #
-# USAGE - define $FILES variable as list of segmentations (separeted by space)
-# to correct in the terminal and then run:
-#   PATH_TO_SPINEGENERIC/processing/manual_correction.sh $FILES
+# USAGE:
+# Define files.yaml file as following:
+#   ---
+#   FILES_SEG:
+#   - sub-amu01_T1w_RPI_r.nii.gz
+#   - sub-amu01_T2w_RPI_r.nii.gz
+#   - sub-brnoCeitec01_T2star_rms.nii.gz
+#   FILES_LABEL:
+#   - sub-amu01
+#   - sub-brnoUhb01
 #
+#
+# THEN RUN:
+#   PATH_TO_SPINEGENERIC/processing/manual_correction.sh files.yaml
+#
+# Authors: Julien Cohen-Adad, Jan Valosek
 
 
-# Print help
+# function for parsing input yaml file
+yaml_parser() {
+
+  INPUT_YAML=$1
+
+  # tr: delete all newlines begining with dash (-), so convert input file into one line without dashes
+  # (unfortunately this also delete all other dashes, so it is neccesary to use 3rd sed)
+  # 1st sed: delete "FILES_LABEL:" string and everything after it
+  # 2nd sed: delete "FILES_SEG:" string itself
+  # 3rd sed: replace "sub" by "sub-" (tr command deleted all dashes (-))
+  FILES_SEG=$(cat $INPUT_YAML | tr -d '\n-' | sed 's/FILES_LABEL:.*//' | sed 's/FILES_SEG://' | sed 's/sub/sub-/g')
+  # tr: delete all newlines begining with dash (-), so convert input file into one line without dashes
+  # (unfortunately this also delete all other dashes, so it is neccesary to use 3rd sed)
+  # 1st sed: delete everything before "FILES_LABEL:" string including string "FILES_LABEL:" itself
+  # 2nd sed: replace "sub" by "sub-" (tr command deleted all dashes (-))
+  FILES_LABEL=$(cat $INPUT_YAML | tr -d '\n-' | sed 's/.*FILES_LABEL://' | sed 's/sub/sub-/g')
+}
+
+# Print help, if invalid input
 if [[ $# != 1 ]] || [[ $1 == "--help" ]] || [[ $1 == "-h" ]];then
 
-  echo "Invalid input. Please define \$FILES variable as list of segmentations (separeted by space) to correct in the terminal."
-  echo -e "Example:\n\tFILES=(sub-amu02_acq-T1w_MTS.nii.gz sub-beijingGE04_T2w_RPI_r.nii.gz)"
-  echo -e "Then run:\n\tPATH_TO_SPINEGENERIC/processing/manual_correction.sh \$FILES"
+  echo -e "Invalid input. \n\nPlease define files.yaml file with list of segmentations to correct"
+  echo -e "and list of subjects to correct labeling on (one file perline)."
+  echo -e "\nExample:"
+  echo -e "\t---"
+  echo -e "\tFILES_SEG:"
+  echo -e "\t- sub-amu01_T1w_RPI_r.nii.gz"
+  echo -e "\t- sub-amu01_T2w_RPI_r.nii.gz"
+  echo -e "\t- sub-brnoCeitec01_T2star_rms.nii.gz"
+  echo -e "\tFILES_LABEL:"
+  echo -e "\t- sub-amu01"
+  echo -e "\t- sub-brnoUhb01"
+  echo -e "\nThen run this script in results/data folder:\n\tPATH_TO_SPINEGENERIC/processing/manual_correction.sh files.yaml"
 
 else
 
-  # Folder to output the manual labels
+  # call yaml_parser function
+  yaml_parser $1
+
+  # Folder to output the manual segmentations and labels
   PATH_SEGMANUAL="../../seg_manual"
   mkdir $PATH_SEGMANUAL
-  # List of files to correct segmentation on
-  FILES=$1
 
-  # Loop across files
-  for file in ${FILES[@]}; do
+  # Loop across segmentation files
+  for file in $FILES_SEG; do
+
     # extract subject using first delimiter '_'
     subject=${file%%_*}
     # check if file is under dwi/ or anat/ folder and get fname_data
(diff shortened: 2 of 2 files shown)


valosekj:
sure, updated!

docs/documentation.rst
```diff
@@ -298,52 +298,8 @@ If you spot segmentation issues, manually fix them using the procedure described
...
-    fsleyes -yh $fname_data $fname_seg_dest -cm red
-  done
-
-- Make this script executable:
-
-  .. code-block:: bash
-
-     chmod 775 fix_seg.sh
-
+- Edit list of files to correct segmentation on, i.e. edit ``FILES`` variable in ``manual_correction.sh`` script
```
jcohenadad, line 301: @valosekj if we add this script in the git repository, it would be more elegant to _not_ have the user edit this file (otherwise it will create conflicts with the origin/master branch). Instead, how about you slightly modify the script to accept `FILES` as an input argument. So the procedure would be:
- in the terminal, define the list of files to correct, example:
~~~
FILES=(
sub-amu02_acq-T1w_MTS.nii.gz
sub-beijingGE04_T2w_RPI_r.nii.gz
sub-brnoPrisma01_T2star_rms.nii.gz
sub-geneva04_dwi_crop_moco_dwi_mean.nii.gz
)
~~~
- Then, call the script with the list of files as input argument:
~~~
PATH_TO_SPINEGENERIC/processing/manual_correction.sh $FILES
~~~

processing/manual_correction.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
jcohenadad, line 1: we should also create a script for the vertebral labeling, or even better: include manual labeling inside this script. You can figure out if the input file is designed as a segmentation or a labeling, by looking at the end of the file (seg vs. label)

docs/documentation.rst
valosekj: Absolutely! I was thinking about passing of this variable as input argument as well.

processing/manual_correction.sh
valosekj: Okay, I will implement it.

docs/documentation.rst
valosekj: Implemented :+1: , if you agree with it, I will update doc.

processing/manual_correction.sh
valosekj: One question before I start with implementation - `$FILES` variable now contains filenames of background images, not filenames of segmentation (or labeling), so I cannot figure out if to open segmentation or labeling. So, do you agree with passing filenames of segmentation/labeling into `$FILES` variable instead of current background image filenames?

jcohenadad: > One question before I start with implementation - `$FILES` variable now contains filenames of background images, not filenames of segmentation (or labeling), so I cannot figure out if to open segmentation or labeling. So, do you agree with passing filenames of segmentation/labeling into `$FILES` variable instead of current background image filenames?

great point! so, i see 2 choices:
- Choice 1: create two variables: $FILES_SEG and $FILES_LABEL
  - Pros: easy copy/paste from the QC window (ie: no need to manually type "_seg" or "_label", which is prone to error)
  - Pros: no need to parse the file to retrieve the file name of the image
  - Pros: if people do the QC of segs, and then of labeling, it might be easier to list the files into two separate variables 
  - Pros: makes code cleaner (i.e. we would have two loops: one for $FILES_SEG that calls FSLeyes, and one for $FILES_LABEL that calls sct_label_utils)
- Choice 2: create a single $FILES, that list the segmentations or labels to modify
  - Pros: might be easier for the user
  - Pros: more intuitive (user points the problematic segmentation/labels), however does require manual modification of file names (humans make mistakes)
  - Cons: code less clean (you will need to parse the file, figure out if it is a seg or a label, and then send to the appropriate if case)

valosekj: Thanks for brainstorming! So, I would vote for the first choice. It eliminates manual adding of "_seg" or "_label" suffix which I can see as the biggest disadvantage and possible prone to errors. I'll implement choice 1.

valosekj: Hm, I have encountered on problem when I tried to pass two list (array in language of bash) variables (`$FILES_SEG` and `$FILES_LABEL`) as inputs to bash script. It looks that bash does not allow to pass arrays, see [here](https://stackoverflow.com/questions/16461656/how-to-pass-array-as-an-argument-to-a-function-in-bash) and [here](https://askubuntu.com/questions/674333/how-to-pass-an-array-as-function-argument).

Workaround could be to pass only one list/array variable as input (so, the choice 2 from previous brainstorming) and use something like this `INPUTS=$@` inside the script to catch all elements of input array. But still here would be disadvantage that script would have to be called `manual_correction.sh ${FILES_SEG[@]}` instead of simply `manual_correction.sh $FILES_SEG`.

Another workaround could be to define input variables as simple strings, pass them to script and split them into individual filenames inside the script.

jcohenadad: how about simply passing a json and/or yaml file? another pros is that the information is saved

valosekj: It is another option. Or simply planar text file with individual filenames as rows. This file would have to be added to .gitignore to do not conflict if user save it into repo folder, however (probably not problem). But sure, saving of this information is huge pros.

jcohenadad: > It is another option. Or simply planar text file with individual filenames as rows. 

no, this is not following proper convention. This is why json and yaml formats exist (universal convention).

> This file would have to be added to .gitignore to do not conflict if user save it into repo folder

personally i've been launching those script from within the output folder (not from the repos).

but yes, people can add it to their .gitignore if they want to

jcohenadad: p.s. yaml is way friendlier than json

valosekj: Okay, yaml looks fine. Which exact yaml format we will use?
inline with comments and without keys
```
---
# FILES_SEG
[sub-amu01_T1w_RPI_r.nii.gz, sub-amu01_T1w_RP_r.nii, sub-amu01_T2w_RPI_r.nii.gz]
# FILES_LABEL
[sub-amu01, sub-amu02]
```
or inline with keys

```
---
FILES_SEG: [sub-amu01_T1w_RPI_r.nii.gz, sub-amu01_T1w_RP_r.nii, sub-amu01_T2w_RPI_r.nii.gz]
FILES_LABEL: [sub-amu01, sub-amu02]
```

jcohenadad: We probably want the key as it will be used for choosing the right pipeline, no?
Also, I’d show one subject per line (PEP8)

valosekj: Yeah, option with keys is probably more cleaner. However, I was able to identify pipeline also based on comment above certain line.
So, yaml file with keys and one subject perline would look:

```
---
FILES_SEG:
- sub-amu01_T1w_RPI_r.nii.gz
- sub-amu01_T1w_RP_r.nii
- sub-amu01_T2w_RPI_r.nii.gz
FILES_LABEL:
- sub-amu01
- sub-amu02
```

jcohenadad: beautiful! now we just need a yaml parser 😊

valosekj: Added here https://github.com/sct-pipeline/spine-generic/pull/87/commits/58e4224f883a89f7e852002a1023689557fe36f3 😊

processing/manual_correction.sh
```diff
@@ -26,13 +26,13 @@ yaml_parser() {
   INPUT_YAML=$1
 
   # tr: delete all newlines begining with dash (-), so convert input file into one line without dashes
-  # (unfortunately this also delete all other dashes, so it is neccesary to use 3rd sed)
+  # (this also deletes all other dashes, so it is necessary to use last sed)
   # 1st sed: delete "FILES_LABEL:" string and everything after it
   # 2nd sed: delete "FILES_SEG:" string itself
   # 3rd sed: replace "sub" by "sub-" (tr command deleted all dashes (-))
```
jcohenadad, line 32: @valosekj if you have troubles with the first "-" in the YAML format we chose, why don't you go with the other YAML:
~~~
# FIELD 1
value1
value2
value3
~~~
does that make sense?

jcohenadad: ^ scrap that comment above: i just realized that "#" is considered a comment, not a key, so that wouldn't be proper use of the YAML format (in case in the future we implement a proper yaml parser).

processing/manual_correction.sh
```diff
@@ -0,0 +1,114 @@
...
+  # 1st sed: delete everything before "FILES_LABEL:" string including string "FILES_LABEL:" itself
+  # 2nd sed: replace "sub" by "sub-" (tr command deleted all dashes (-))
+  FILES_LABEL=$(cat $INPUT_YAML | tr -d '\n-' | sed 's/.*FILES_LABEL://' | sed 's/sub/sub-/g')
+}
+
+# Print help, if invalid input
+if [[ $# != 1 ]] || [[ $1 == "--help" ]] || [[ $1 == "-h" ]];then
+
+  echo -e "Invalid input. \n\nPlease define files.yaml file with list of segmentations to correct"
+  echo -e "and list of subjects to correct labeling on (one file perline)."
```
jcohenadad, line 45: quick search shows that yml is more common
also, replace the "here" by an example. people need examples
```suggestion
  echo -e "Invalid input. \n\nPlease create a files.yml file, which lists the files associated with the segmentations or labels to manually correct"
```
```suggestion
  echo -e "Invalid input. \n\nPlease define files.yaml file with list of segmentations to correct"
  echo -e "and list of subjects to correct. See example HERE"
```

processing/manual_correction.sh
```diff
@@ -0,0 +1,114 @@
...
+  echo -e "\nExample:"
+  echo -e "\t---"
```
jcohenadad, line 47: not needed, i think

processing/manual_correction.sh
```diff
@@ -0,0 +1,114 @@
...
+  echo -e "\tFILES_SEG:"
```
jcohenadad, line 48: why the \t ?

processing/manual_correction.sh
```diff
@@ -0,0 +1,114 @@
...
+  echo -e "\t- sub-brnoUhb01"
+  echo -e "\nThen run this script in results/data folder:\n\tPATH_TO_SPINEGENERIC/processing/manual_correction.sh files.yaml"
+
+else
+
+  # call yaml_parser function
+  yaml_parser $1
+
+  # Folder to output the manual segmentations and labels
+  PATH_SEGMANUAL="../../seg_manual"
```
jcohenadad, line 63: strong assumption-- this should at least be documented somewhere-- or could be input arg

processing/manual_correction.sh
valosekj: Yes, "#" denotes comments. I would let current format which follows [yaml syntax](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html). Originally, I wanted to use `sed` command for removing `newline` character and first `-`, but i found out that default `sed` on OSX does not support `\n`. So i had to workaround it using `tr` command.

processing/manual_correction.sh
valosekj: For better looking when printed as help into the terminal.

processing/manual_correction.sh
valosekj: So, delete whole "help" from bash script?

jcohenadad: No I meant the “——“

processing/manual_correction.sh
valosekj: But it is true it can confuse users. I deleted it.

processing/manual_correction.sh
valosekj: You are right 👍 I have mentioned it in the script description. But sure, it could be input arg. We could you e.g. `getopts` built-in bash function ([example](https://sookocheff.com/post/bash/parsing-bash-script-arguments-with-shopts/)).

processing/manual_correction.sh
valosekj: >(in case in the future we implement a proper yaml parser).

We could use some of already created yaml parsers (e.g., [here](https://github.com/jasperes/bash-yaml)). On the other hand, if we don't modify structure of yaml file, currently proposed simple dedicated parser could be sufficient.

Mentioned also in https://github.com/sct-pipeline/csa-atrophy/issues/23

jcohenadad:
@valosekj your branch is not up-to-date with master

docs/documentation.rst
jcohenadad: @valosekj i thought we are using a yml file now? the doc should be updated

processing/manual_correction.sh
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
+
+# Run this script in results/data folder
+#
+# Manual segmentations and labels are saved into ../../seg_manual folder
+# (feel free to change this folder by change PATH_SEGMANUAL variable)
```
jcohenadad, line 6: where is this PATH_SEGMANUAL variable defined? it's confusing

processing/manual_correction.sh
```diff
@@ -0,0 +1,116 @@
...
+#
+# USAGE:
+# Define files.yml file as following:
```
jcohenadad, line 9:
```suggestion
# Create a files.yml file that list the files to manually correct, for the segmentation (FILES_SEG) and vertebral labeling (FILES_LABEL), as per the example below:
```

processing/manual_correction.sh
```diff
@@ -0,0 +1,116 @@
...
+#   FILES_SEG:
+#   - sub-amu01_T1w_RPI_r.nii.gz
+#   - sub-amu01_T2w_RPI_r.nii.gz
+#   - sub-brnoCeitec01_T2star_rms.nii.gz
+#   FILES_LABEL:
+#   - sub-amu01
```
jcohenadad, line 15: these are not files but sites

processing/manual_correction.sh
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
+
+# Run this script in results/data folder
+#
+# Manual segmentations and labels are saved into ../../seg_manual folder
+# (feel free to change this folder by change PATH_SEGMANUAL variable)
+#
+# USAGE:
+# Define files.yml file as following:
+#   FILES_SEG:
```
jcohenadad, line 10: we have not distinguished GM from SC seg... we probably should

processing/manual_correction.sh
jcohenadad: actually, why not making it a python script? advantages:
- can be added as entry point script in the setup.py so could be called from anywhere
- python is friendlier (better argparse, better yaml parser, easier debugging, etc.)
- more harmonious for the package (we already have a few python scripts)

valosekj: Good idea. Many of mentioned problems will be easier to solve. I will start with building of it. I will create new PR when draft will be ready.

valosekj:
This PR can be discontinued and then deleted. Replaced by https://github.com/spine-generic/spine-generic/pull/103

jcohenadad:
> This PR can be discontinued and then deleted. Replaced by #103

context: the reason we close this PR without merging is because we decided to replace a sh script by a python script.

processing/manual_correction.sh
valosekj: Was done based on t2star contrast:

https://github.com/spine-generic/spine-generic/blob/957bf91d14250c26527f2b9debea821824fbad8d/processing/manual_correction.sh#L79

Do you think that separate keys `FILE_SEG_SC` and `FILE_SEG_GM`, how is proposed here are better?:
 https://github.com/spine-generic/spine-generic/blob/3b7fbeccc3e5b891b5271b896bb70d557ba471cb/processing/manual_correction.py#L11

For me, it is more understandable at the first look because user explicitly defines SC or GM.

jcohenadad: > Was done based on t2star contrast:

ok all good then

Issue #105, opened 2020-07-03
Upload manual corrections under derivatives/


jcohenadad:
Related to https://github.com/spine-generic/spine-generic/pull/104

- [x] Upload to data-singe → done in https://github.com/spine-generic/data-single-subject/commit/953192ef94cd934bfc7615c7a2d9dde9502e9e62
- [x] Upload to data-multi

PR #106, opened 2020-07-03
New script to copy manual corrections under derivatives/

jcohenadad:
This PR introduces a util script to copy manual corrections placed in a flat directory, under the BIDS_DATA/derivatives/ folder as per BIDS convention.
 
Related to #105


valosekj:
Great 👍 Just tested and works fine. User only have to pay attention to use Python > 3.2 where `exist_ok` was introduced for `os.makedirs` (but probably, it is a matter of course).

PR #115, opened 2020-07-08
Ensures reproducible pipeline across OSs; various improvements

### LINKED ISSUE #110: Non-reproducible results across OSs, causing incompatibilities with the produced manual segmentation
Because propseg results are slightly different across OSs (for more details, see https://github.com/neuropoly/spinalcordtoolbox/issues/2769), when running the pipeline on one OS, correcting a mask, and then re-running in another OS, the following bug could occur, when computing DTI metrics with dipy:
````
+ sct_dmri_compute_dti -i sub-tokyoSigna2_dwi_crop_moco.nii.gz -bvec sub-tokyoSigna2_dwi.bvec -bval sub-tokyoSigna2_dwi.bval -method standard -m sub-tokyoSigna2_dwi_crop_moco_dwi_mean_seg_dil.nii.gz

--
Spinal Cord Toolbox (git-cfh/sct-run-batch-improvements_2726-3fdbf4fb40879240260a9af429bf4dcbd3ee0375)

data.shape (33, 36, 15, 35)
Open mask file...
Computing tensor using "standard" method...
Traceback (most recent call last):
  File "/home/jcohen/code/sct/scripts/sct_dmri_compute_dti.py", line 208, in <module>
    main()
  File "/home/jcohen/code/sct/scripts/sct_dmri_compute_dti.py", line 120, in main
    if not compute_dti(fname_in, fname_bvals, fname_bvecs, prefix, method, evecs, file_mask):
  File "/home/jcohen/code/sct/scripts/sct_dmri_compute_dti.py", line 164, in compute_dti
    tenfit = tenmodel.fit(data, mask)
  File "/home/jcohen/code/sct/python/envs/venv_sct/lib/python3.6/site-packages/dipy/reconst/dti.py", line 777, in fit
    raise ValueError("Mask is not the same shape as data.")
ValueError: Mask is not the same shape as data.
````

A workaround would be to regrid the mask to the DWI before running DTI fit.

jcohenadad:
Because propseg results are slightly different across OSs (for more details, see neuropoly/spinalcordtoolbox#2769), the bug in #110 could occur. 

This PR fixes the problem by:
- replacing propseg by get_centerline

Fixes #110

### DIFF
--- a/processing/process_data.sh
+++ b/processing/process_data.sh
@@ -255,15 +255,13 @@ file_bval=${file_dwi}.bval
 file_bvec=${file_dwi}.bvec
 # Separate b=0 and DW images
 sct_dmri_separate_b0_and_dwi -i ${file_dwi}.nii.gz -bvec ${file_bvec}
-# Segment cord (1st pass -- just to get a rough centerline)
-sct_propseg -i ${file_dwi}_dwi_mean.nii.gz -c dwi
+# Get centerline
+sct_get_centerline -i ${file_dwi}_dwi_mean.nii.gz -c dwi
 # Create mask to help motion correction and for faster processing
-sct_create_mask -i ${file_dwi}_dwi_mean.nii.gz -p centerline,${file_dwi}_dwi_mean_seg.nii.gz -size 30mm
-# Crop data for faster processing
-sct_crop_image -i ${file_dwi}.nii.gz -m mask_${file_dwi}_dwi_mean.nii.gz -o ${file_dwi}_crop.nii.gz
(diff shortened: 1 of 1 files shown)


valosekj:
Sorry for delay. I have been testing it on one subject and waiting for finish. Works fine 👍 

Only; isn't usage obsolete?
https://github.com/spine-generic/spine-generic/blob/033e7f76ef46d40afacbeba5025c863e4adf4c81/processing/process_data.sh#L6

I tested it:
`sct_run_batch -jobs -1 -path-data /md2/ds002900-download/ -path-output /md2/ds002900-download_results/ -include sub-douglas processing/process_data.sh`

jcohenadad:
> Only; isn't usage obsolete?

i don't think it's obsolete, as this is the argument passed by sct_run_batch wrapper-- some things will break (like all the other environment variables), but the usage remains the same.

PR #118, opened 2020-07-09
Adding script to copy label file into derivatives/labels/sub-XXX/anat

lrouhier:
This PR aims at creating a script that copies files from sub-xxx/anat to derivatives/labels/sub-xxx/anat. 
For now, the script is called as follow:
move_file_SG -p path/to/results/file -s suffix_of_file
 
the script: 
* go to path/results/ 
* List all object in folder 
* Check if ./sub-xxx/anat/sub-xxx_SUFFIX.nii.gz exist
* Create directory derivatives/labels/sub-xxx/anat
* copy  sub-xxx/anat/sub-xxx_SUFFIX.nii.gz to  derivatives/labels/sub-xxx/anat
* return the number of moved file (just to check if there is no error) 

pros and cons of asking for fully file suffix: 
* cons:
You need to call it once for each contrast

* pros:
you don't have to test if all contrasts are present.


jcohenadad:
@lrouhier ready for review?

lrouhier:
Just a question I thought about this morning: In the results how is the structure? 
is it sub-xxx/anat/files ?

jcohenadad:
> Just a question I thought about this morning: In the results how is the structure?
> is it sub-xxx/anat/files ?

yup! and for dwi: sub-xxx/dwi/files

lrouhier:
Alright I'll do something more modulable for dwi then !

jcohenadad:
ah! no wait, i think we only need vertebral labels, right? these are only present in anat

jcohenadad:
to be more precise: afair, for Paul's project specifically, what we need is: sub-XXX_T1w_RPI_r_seg_labeled.nii.gz

lrouhier:
So for now the usage would be ./move_files_derivatives.sh SUFFIX of file so for Paul's project 
`./move_files_derivatives.sh T1w_RPI_r_seg_labeled`
Should we keep it that way or hardcode the file suffix?

jcohenadad:
> So for now the usage would be ./move_files_derivatives.sh SUFFIX of file so for Paul's project
> `./move_files_derivatives.sh T1w_RPI_r_seg_labeled`
> Should we keep it that way or hardcode the file suffix?

sounds good to me 👍 
however, could you pls make it a .py script?  i envision this folder to become a proper python package, with callable CLI scripts (and proper setup, etc.).

lrouhier:
I get it. I can do it! will @ you when done.

lrouhier:
@jcohenadad I think it is good. 

To call the CLI move_file_SG -p path/to/results -s suffix_of_file. 
I Updated the body of the PR with the explanation on how to use it. I don't know if the path argument is relevant  ?

processing/move_file_derivatives.py
```diff
@@ -0,0 +1,40 @@
+import os
```
jcohenadad, line 1: can you please add a header to briefly explain what this function does

lrouhier: Done in 48edbd7

jcohenadad:
Note: currently the CLI call fails because utils cannot be imported. This will be fixed via #96

PR #122, opened 2020-07-10
Documentation updates

valosekj:
This PR includes several fixes for the documentation:
- [x] Update documentation for installation, based on changes in #121
- [x] Update usage  of `generate_figure.py` (remove `parameters.sh` input)
- [x] Fix link to `config_spine.txt` file
- [x] Update segmentation and labeling instructions - introduce `manual_correction.py` script and yaml file example
- [x] Update how to exclude subject with low data quality from analysis 
- [x] Add `sg_` prefix to py scripts
- [x] Fix path to `process_data.sh` script (from `processing/process_data.sh` to `process_data.sh`) 
- [x] Document: New script to package data awaiting correction #129
- [x] Fix displayed version

Fixes https://github.com/spine-generic/spine-generic/issues/112

PR #129, opened 2020-07-13
New script to package data awaiting correction

### LINKED ISSUE #128: Create a script that creates a package for the files that need correction
In case processing is ran on a remote cluster, it would be convenient to generate a package of the files that need correction, to conveniently copy them locally (instead of copying the ~20GB of processed files). 

That script would:
- take as input the yml file that lists the files to correct
- copy them in a temp folder
- zip the package

jcohenadad:
In case processing is ran on a remote cluster, it would be convenient to generate a package of the files that need correction, to conveniently copy them locally (instead of copying the ~20GB of processed files). 

The proposed script, name `package_for_correction.py`, and called via CLI with `sg_package_for_correction`:
- take as input the yml file that lists the files to correct
- copy them in a temp folder
- zip the package

Fixes #128

TODO in other PRs
- document this new function in https://github.com/spine-generic/spine-generic/pull/122 (to prevent conflicts)

### DIFF
--- a/spinegeneric/cli/package_for_correction.py
+++ b/spinegeneric/cli/package_for_correction.py
@@ -63,8 +63,8 @@ def get_parser():
     parser.add_argument(
         '-o',
         metavar=sg.utils.Metavar.folder,
-        help="Zip file that contains the packaged data. Default: data_to_correct.zip",
-        default='data_to_correct.zip'
+        help="Zip file that contains the packaged data, without the extension. Default: data_to_correct",
+        default='data_to_correct'
     )
     parser.add_argument(
         '-v', '--verbose',
--- a/spinegeneric/cli/package_for_correction.py
+++ b/spinegeneric/cli/package_for_correction.py
@@ -10,6 +10,7 @@
 import os
 import sys
 import shutil
+import tempfile
 from textwrap import dedent
 import argparse
 import yaml
@@ -18,11 +19,6 @@ import coloredlogs
 import spinegeneric as sg
 import spinegeneric.utils
 import spinegeneric.bids
-
-
-# Folder where to output manual labels, at the root of a BIDS dataset.
-# TODO: make it an input argument (with default value)
-FOLDER_DERIVATIVES = os.path.join('derivatives', 'labels')
 
 
 def get_parser():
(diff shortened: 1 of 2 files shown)


spinegeneric/cli/package_for_correction.py
```diff
@@ -0,0 +1,144 @@
...
+        '-path-in',
+        metavar=sg.utils.Metavar.folder,
+        help='Path to the processed data. Example: ~/spine-generic/results/data',
+        default='./'
+    )
+    parser.add_argument(
+        '-o',
+        metavar=sg.utils.Metavar.folder,
+        help="Zip file that contains the packaged data. Default: data_to_correct.zip",
+        default='data_to_correct.zip'
```
valosekj, line 67: I had to delete suffix `.zip` at this line (to` default='data_to_correct'`). Otherwise, output file is named `some_name.zip.zip` because line 139 add zip suffix as well.

valosekj:
When print help of this script, first letter `p` in script name is missing:

image

spinegeneric/cli/package_for_correction.py
```diff
@@ -0,0 +1,144 @@
...
+        description='Package data for manual correction. In case processing is ran on a remote cluster, it it '
+                    'convenient to generate a package of the files that need correction to be able to only copy these '
+                    'files locally, instead of copying the ~20GB of total processed files.',
+        formatter_class=sg.utils.SmartFormatter,
+        prog=os.path.basename(__file__).strip('.py')
+    )
+    parser.add_argument(
+        '-config',
+        metavar=sg.utils.Metavar.file,
+        required=True,
```
valosekj, line 38: Although is this argument required, when I print help, it is listed under `optional arguments`. Maybe, it could be confusing for somebody.

image

spinegeneric/cli/package_for_correction.py
jcohenadad: great catch! fixed in 01069533b7dcf44211faa356bb0e7833713b1a55

jcohenadad:
> When print help of this script, the first letter `p` in script name is missing:
> 
> image

fixed in d8ecb9d5bc05a7bea113477a5d34d23a706b5bc9

jcohenadad:
> When print help of this script, the first letter `p` in script name is missing:
> 
> image

this is the downside of having non-positional arguments (which i prefer, in general). If users forget to specify -config, they get the explicit answer: 
~~~
package_for_correction: error: the following arguments are required: -config
~~~

jcohenadad:
@valosekj good to merge?

PR #146, opened 2020-07-22
Generate QC report for manual corrections

### LINKED ISSUE #145: Produce QC report after manual correction
It would be nice to generate a QC report along with manual corrections, so they can be easily reviewed by others. 

The idea would be to zip that QC folder and upload it along with a new release of the updated database.

jcohenadad:
In this PR, we introduce the generation of a QC report of manual corrections, which can be uploaded alongside a new release of the spine-generic dataset.  Specific changes include:
- QC report for segmentations and label
- QC report generated locally
- QC report archived as zip file for convenient upload alongside data release on github
- Updated documentation on RTD

Fixes #145

### DIFF
--- a/docs/analysis_pipeline.rst
+++ b/docs/analysis_pipeline.rst
@@ -153,6 +153,13 @@ and open an interactive window for you to either correct the segmentation, or pe
 manually-corrected label is saved under the ``derivatives/labels/`` folder at the root of ``<PATH_DATA>``,
 according to the BIDS convention. The manually-corrected label files have the suffix ``-manual``.
 
+Your name will be asked at the beginning, and will be recorded in the .json files that accompany the corrected labels.
+
+A QC report of all the manual correction will be created locally and archived as a zip file. To update the
+database with the manual corrections (using git), create a pull request and upload the QC report directly in the pull
+request so the admin team can easily review the proposed changes. If the team accepts the pull request, a new release
+of the dataset will be created and the zipped QC report will be uploaded as a release object.
+
 .. note::
 
    In case processing is ran on a remote cluster, it it convenient to generate a package of the files that need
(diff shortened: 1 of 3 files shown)


jcohenadad:
@valosekj and @PaulBautin can you give it a try?

valosekj:
@jcohenadad works good 👍 

Only two things:

1. where do we want to save QC report? In the current implementation it is saved into working directory from which the script `sg_manual_correction.py` is run. Maybe, the path for saving QC report could be constructed inside the script for the situation when user runs the script e.g. from home directory with manually specification of` -path-in` and `-path-out` flags (then QC would be saved into home directory).
2. a lot of stuff is printed into terminal when QC is creating. Is it expected? Can't it be disruptive for user?
image

jcohenadad:
> In the current implementation it is saved into working directory from which the script sg_manual_correction.py is run. 

are you sure about that? it is supposed to be created in the local folder (pwd), which is intuitive and convenient because users will have to manually upload it on the web afterwards.

> a lot of stuff is printed into terminal when QC is creating. Is it expected? Can't it be disruptive for user? On the other hand, user knows what is happening.

it’s important verbose information. i think it should be shown in default behaviour

valosekj:
> > In the current implementation it is saved into working directory from which the script sg_manual_correction.py is run.
> 
> are you sure about that? it is supposed to be created in the local folder (pwd), which is intuitive and convenient because users will have to manually upload it on the web afterwards.

Sorry, I wasn't correct. I meant pwd.
I have just read updated documentation about upload. Then, you are right, pwd is intuitive.

> 
> > a lot of stuff is printed into terminal when QC is creating. Is it expected? Can't it be disruptive for user? On the other hand, user knows what is happening.
> 
> it’s important verbose information. i think it should be shown in default behaviour

Okay, sounds good.

PR #164, opened 2020-08-04
fix superscript for MD and RD ylabel and remove a.u. unit for MTsat

### LINKED ISSUE #163: sg_generate_figure ylabel cosmetic issues
I have found two cosmetic issues when plotting figures using `sg_generate_figure` script:

1. Last digit in labels on y-axis in mean and radial diffusivity (MD and RD) figures is not in superscript:

image

2. Magnetic transfer saturation has `a.u.` as an unit, but fractional anisotropy has no unit. Don't we want to unify this, i.e., add `a.u.` also for FA or delete `a.u.` from MTsat label?

image

image

I could fix this.

valosekj:
This PR fixes two cosmetic issues in figures generated by `sg_generate_figure` script:

- fix superscript in MD and RD ylabel
- change a.u. unit for MTsat to [%]

Fixes #163

### DIFF
--- a/spinegeneric/cli/generate_figure.py
+++ b/spinegeneric/cli/generate_figure.py
@@ -170,10 +170,10 @@ metric_to_label = {
     'csa_t2': 'Cord CSA from T2w [$mm^2$]',
     'csa_gm': 'Gray Matter CSA [$mm^2$]',
     'dti_fa': 'Fractional anisotropy',
-    'dti_md': 'Mean diffusivity [$mm^2.s^-1$]',
-    'dti_rd': 'Radial diffusivity [$mm^2.s^-1$]',
+    'dti_md': 'Mean diffusivity [$mm^2.s^{-1}$]',
+    'dti_rd': 'Radial diffusivity [$mm^2.s^{-1}$]',
     'mtr': 'Magnetization transfer ratio [%]',
-    'mtsat': 'Magnetization transfer saturation [a.u.]',
+    'mtsat': 'Magnetization transfer saturation',
     't1': 'T1 [ms]',
     }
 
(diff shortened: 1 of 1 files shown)


spinegeneric/cli/generate_figure.py
```diff
@@ -170,10 +170,10 @@
     'csa_t2': 'Cord CSA from T2w [$mm^2$]',
     'csa_gm': 'Gray Matter CSA [$mm^2$]',
     'dti_fa': 'Fractional anisotropy',
-    'dti_md': 'Mean diffusivity [$mm^2.s^-1$]',
-    'dti_rd': 'Radial diffusivity [$mm^2.s^-1$]',
+    'dti_md': 'Mean diffusivity [$mm^2.s^{-1}$]',
+    'dti_rd': 'Radial diffusivity [$mm^2.s^{-1}$]',
     'mtr': 'Magnetization transfer ratio [%]',
-    'mtsat': 'Magnetization transfer saturation [a.u.]',
+    'mtsat': 'Magnetization transfer saturation',
```
jcohenadad, line 176: i think it is actually a percentage (correct me if i'm wrong)
```suggestion
    'mtsat': 'Magnetization transfer saturation [%]',
```

valosekj: You are right 👍 
In the `compute_mtsat` function [docstring](https://github.com/neuropoly/spinalcordtoolbox/blob/3da4b2697093d45969691fd3e48d1a2f85ddc197/spinalcordtoolbox/qmri/mt.py#L34 ) is written: "Compute MTsat (in percent) and T1 map (in s) based on FLASH scans"
and percentage are also stated in original Helms et el., 2008 paper:

image

Issue #165, opened 2020-08-11
Rename sapienza --> fslAchieva


jcohenadad:
After merging https://github.com/spine-generic/data-multi-subject/pull/21, we should rename here (and at other places):

https://github.com/spine-generic/spine-generic/blob/1329c37e40569ab22c37526d2d3ac9eb2b8f23a6/spinegeneric/cli/generate_figure.py#L60

alexfoias:
I can work on this.

alexfoias:
@jcohenadad Can I also add fslPrisma in the flags sections ? or should I open another PR for that ?

jcohenadad:
> @jcohenadad Can I also add fslPrisma in the flags sections ?

👍

PR #167, opened 2020-08-11
Fix generate figure

alexfoias:
- Rename sapienza --> fslAchieva
- Add flag for fslPrisma

Fixes https://github.com/spine-generic/spine-generic/issues/165

### DIFF
--- a/spinegeneric/cli/generate_figure.py
+++ b/spinegeneric/cli/generate_figure.py
@@ -112,6 +112,7 @@ flags = {
     'poly': 'canada',
     'queensland': 'australia',
     'fslAchieva': 'italy',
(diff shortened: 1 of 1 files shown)


alexfoias:
@jcohenadad generate_figure will fail with the dataset sct_run_batch_20200722022726/results. the dataset has to be reprocessed or change the mange of the center in the results

jcohenadad:
> @jcohenadad generate_figure will fail with the dataset sct_run_batch_20200722022726/results

which dataset is that? are you referring to a release?

> the dataset has to be reprocessed or change the mange of the center in the results

sorry i don't understand the sentence (mange?)

PR #179, opened 2020-08-27
generate_figure: small improvements; code clarifications

### LINKED ISSUE #178: generate_figure: Subjects to exclude are counted in the statistics for cov_intra calculation
They are just removed from the plot, but counted in the stats. This needs to be fixed.

This function call:
https://github.com/spine-generic/spine-generic/blob/8b014c68fc721314b6d395d0eb8e635ffd726249/spinegeneric/cli/generate_figure.py#L514

Takes in argument the sites (not subjects) to exclude.

So, one strategy, for a clean code, would be to have two separate dataframes:
- one for plotting
- one for statistics

Then, depending on what user is asking (exclude from plotting or stats or both), we can drop rows in the appropriate dataframe(s).

### LINKED ISSUE #180: COV_intra should not be displayed for single-subject case
For the single-subject data, the generated plots should avoid showing the COV intra site, which does not make sense:

image

jcohenadad:
This PR includes the following improvements:
- Exclude outlier sites from cov_intra computation. Fixes #178
- Do not display COV_intra for single-subject case. Fixes #180
- Display red cross if site is excluded from statistics.
- Adjust figure size horizontally based on number of sites.
- Adjust ylim, improved dynamic (ylim[0] is no more 0)
- Refactoring to nest functions and simplify code.

TODO:
- change flag: `-indiv-subj` --> `-indiv-sub` and document

### DIFF
--- a/spinegeneric/cli/generate_figure.py
+++ b/spinegeneric/cli/generate_figure.py
@@ -412,9 +412,6 @@ def generate_figure_metric(df, metric, stats, display_individual_subjects):
             val = [value * scaling_factor.get(metric) for value in val]
             plt.plot([index] * len(val), val, 'r.')
 
-    # Add ManufacturersModelName embedded in each bar
-    ax = label_bar_model(ax, bar_plot, model_sorted)
-
     # Deal with xticklabels
     # Rotate xticklabels at 45deg, align at end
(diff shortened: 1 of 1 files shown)


valosekj (approved):
Tested and works fine 👍

PR #182, opened 2020-08-28
Replaced deepseg by propseg

### LINKED ISSUE #181: Large variability of CSA results on single-subject data
Compared to the initial results from 2017, the segmentation with deepseg seems to yield more variability.

From https://github.com/spine-generic/data-single-subject/releases/tag/r20200826:
fig_csa_t1

It would be interesting to compare, with the same codebase and dataset, propseg and deepseg results.

jcohenadad:
Fixes #181


jcohenadad:
at the end, deepseg provided best COVs

PR #192, opened 2020-09-04
Removed shell-based fetching of BIDS metadata for MTsat

jcohenadad:
There was a bug for fetching the TR and FlipAngle from the json metadata in the current version of `process_data`, producing wrong results. See e.g. https://github.com/spine-generic/data-multi-subject/issues/44#issuecomment-686759403.

There is now a feature in SCT' `sct_compute_mtsat`, where TR and FA are automatically retrieved if the json sidecar is present. More details in https://github.com/neuropoly/spinalcordtoolbox/pull/2875.

This PR removes the shell-based fetching of the BIDS metadata and uses SCT's new Python-based feature instead.

### DIFF
--- a/process_data.sh
+++ b/process_data.sh
@@ -52,13 +52,6 @@ concatenate_b0_and_dwi(){
   fi
 }
 
-# Get specific field from json file
-get_field_from_json(){
-  local file="$1"
-  local field="$2"
-  echo `grep $field $file | sed 's/[^0-9]*//g'`
-}
-
 # Check if manual label already exists. If it does, copy it locally. If it does
 # not, perform labeling.
 label_if_does_not_exist(){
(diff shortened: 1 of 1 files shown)


process_data.sh
```diff
@@ -223,6 +209,10 @@ if [[ -e "${file_t1w}.nii.gz" && -e "${file_mton}.nii.gz" && -e "${file_mtoff}.n
   # Register MT->T1w
   sct_register_multimodal -i ${file_mton}.nii.gz -d ${file_t1w}.nii.gz -dseg ${file_t1w_seg}.nii.gz -param step=1,type=im,algo=rigid,slicewise=1,metric=CC -x spline -qc ${PATH_QC} -qc-subject ${SUBJECT}
   file_mton="${file_mton}_reg"
+  # Copy json files to match file basename (it will later be used by sct_compute_mtsat)
+  cp ${SUBJECT}_acq-T1w_MTS.json ${file_t1w}.json
+  cp ${SUBJECT}_acq-MToff_MTS.json ${file_mtoff}.json
+  cp ${SUBJECT}_acq-MTon_MTS.json ${file_mton}.json
```
valosekj, line 215: Are these lines really necessary? It looks that e.g. ${file_t1w} variable contains exactly the same filename how the original json file:

https://github.com/spine-generic/spine-generic/blob/1c113421b4006dc65fc067ebdd38eb6344c87631/process_data.sh#L192-L194

valosekj (approved):

process_data.sh
jcohenadad: yes, because file names get updated, eg here:
https://github.com/spine-generic/spine-generic/blob/1c113421b4006dc65fc067ebdd38eb6344c87631/process_data.sh#L204

jcohenadad: yes, because those files are processed, new .nii.gz files are generated, but not new json files, E.g.: 
https://github.com/spine-generic/spine-generic/blob/1c113421b4006dc65fc067ebdd38eb6344c87631/process_data.sh#L204

Issue #193, opened 2020-09-04
generate_figure: ValueError: x and y must be the same size occurs if T1w and T2w subjects do not match


jcohenadad:
If the subjects don't match between T1w and T2w, the following error is raised:
~~~
Traceback (most recent call last):
  File "/Users/julien/miniconda3/bin/sg_generate_figure", line 11, in <module>
    load_entry_point('spinegeneric', 'console_scripts', 'sg_generate_figure')()
  File "/Users/julien/code/spine-generic/spine-generic/spinegeneric/cli/generate_figure.py", line 791, in main
    generate_figure_t1_t2(df, csa_t1, csa_t2)
  File "/Users/julien/code/spine-generic/spine-generic/spinegeneric/cli/generate_figure.py", line 555, in generate_figure_t1_t2
    label=vendor)
  File "/Users/julien/miniconda3/lib/python3.7/site-packages/matplotlib/pyplot.py", line 2841, in scatter
    None else {}), **kwargs)
  File "/Users/julien/miniconda3/lib/python3.7/site-packages/matplotlib/__init__.py", line 1599, in inner
    return func(ax, *map(sanitize_sequence, args), **kwargs)
  File "/Users/julien/miniconda3/lib/python3.7/site-packages/matplotlib/axes/_axes.py", line 4443, in scatter
    raise ValueError("x and y must be the same size")
ValueError: x and y must be the same size
~~~

An elegant solution would be to ignore the subject _both_ for T1w and T2w, in case it is excluded in one of them. 

~Another (quick) solution is to ignore the excluded subjects.~ 

**Update**: oopsi! i tried running the code without the `-exclude` flag and the same error happens 😞

jcohenadad:
> Update: oopsi! i tried running the code without the -exclude flag and the same error happens 😞

actually, this was caused by corrupted csv files with duplicated subjects (process run several times):
[Archive.zip](https://github.com/spine-generic/spine-generic/files/5176109/Archive.zip)

so, ignoring excluded subjects _should_ work.

valosekj:
I have just tried to generate this figure with previous release of results ([r20200830](https://github.com/spine-generic/data-multi-subject/releases/tag/r20200830)) and it fails on the same error. 
[...]

PR #199, opened 2020-09-07
Fix different x and y size in t1-t2 agreement figure

valosekj:
This PR proposed one of possible fix for https://github.com/spine-generic/spine-generic/issues/193.

For successful creation of t1-t2 agreement figure, it is necessary to x and y-axis have same size (i.e., same number of subjects). This is not satisfied when:
- some subjects are excluded from `csa_t1` but not from `csa_t2` (or vice versa)
- no exclude.yml file is used, but some subjects have in input `csa-SC_T1w.csv` or `csa-SC_T2w.csv` files _None_ values

This PR adds:
- merging of subjects from `csa_t1` and `csa_t2` keys. For example, when input `exclude.yml` file looks like:

```
csa_t1:
- sub-oxfordFmrib04  # this is only example
- sub-mountSinai03  # this is only example
csa_t2:
- sub-beijingPrisma03  # this is only example
```

then, after merging inside script, it looks like:

```
csa_t1:
- sub-oxfordFmrib04  # this is only example
- sub-mountSinai03  # this is only example
- sub-beijingPrisma03  # this is only example
csa_t2:
- sub-oxfordFmrib04  # this is only example
- sub-mountSinai03  # this is only example
- sub-beijingPrisma03  # this is only example
```

i.e. `csa_t1` and `csa_t2` are same.

- error message when no exclude.yml file is used and some subject has _None_ values

**CONS** - with this workaround, subjects listed only in `csa_t1` (or only in `csa_t2`) are excluded not only from t1-t2 agreement figure, but also from `fig_csa_t1.png` and `fig_csa_t2.png` bars figures. It is not ideal, because when subject has only bad `fig_csa_t1.png`, it is not necessary to exclude him also from `fig_csa_t2.png`. But, to ensure this, it would require bigger code refactoring because subjects are excluded for each metric (t1, t2,...) separatelly and grouped by sites (so, sub_ID is lost and only lists of values for sites are further processed).

Fixes: https://github.com/spine-generic/spine-generic/issues/193

### DIFF
--- a/spinegeneric/cli/generate_figure.py
+++ b/spinegeneric/cli/generate_figure.py
@@ -523,36 +523,81 @@ def generate_figure_metric(df, metric, stats, display_individual_subjects, show_
     logger.info('Created: ' + fname_fig)
 
 
-def generate_figure_t1_t2(df, csa_t1, csa_t2):
+def generate_figure_t1_t2(dict_exclude_subj):
     """
-    Generate CSA_T1w vs. CSA_T2w
-    :param df:
-    :param csa_t1:
-    :param csa_t2:
+    :param dict_exclude_subj: dict with subjects to exclude
     :return:
     """
 
-    # Sort values per vendor
-    site_sorted = df.sort_values(by=['vendor', 'model', 'site']).index.values
-    vendor_sorted = df['vendor'][site_sorted].values
+    def _build_metric_df(csv_file):
+        """
+        Build pandas DF for csa-SC_T1w.csv and csa-SC_T2w.csv result files
+        :param csv_file: filename of input .csv file
+        :return: metric_df: pandas DF with mean values and subject_ID as indexes
+        """
+        # Read only columns with filename (contains sub_ID) and metric mean value
+        metric = pd.read_csv(csv_file, usecols=['Filename', 'MEAN(area)'])
 
-    # Create dictionary with CSA for T1w and T2w per vendors
-    CSA_dict = defaultdict(list)
-    for index, line in enumerate(csa_t1):       # loop through individual sites
-        CSA_dict[line[1] + '_t1'].append(np.asarray(csa_t1[index, 3]))      # line[1] denotes vendor
-        CSA_dict[line[1] + '_t2'].append(np.asarray(csa_t2[index, 3]))      # line[1] denotes vendor
+        subject_list = list()
+        mean_list = list()
+        # Loop across individual lines in metric DF (i.e., across subjects)
+        for index, line in metric.iterrows():
+            # Fetch sub_ID based on metric filename
+            subject_list.append(fetch_subject(line['Filename']))
+            # Get mean value
+            mean_list.append(line['MEAN(area)'])
+
+        # Create pandas DF from mean_list with subject_list as indexes
+        metric_df = pd.DataFrame(mean_list, index=subject_list, columns=[csv_file[:-4]])
+        return metric_df
+
+    logger.info('\nGenerating T1w and T2w agreement figures.')
+
+    # Build pandas DF of participants based on participants.tsv file (this file contains info about sites and vendors
+    # for individual subjects)
+    participants = pd.read_csv(os.path.join('participants.tsv'), sep="\t")
+    # Create pandas DF with from vendors as subjects_ID as indexes
+    vendor_pd = pd.DataFrame(list(participants['manufacturer']), index=participants['participant_id'],
+                             columns=['vendor'])
+
+    # Fetch csa-SC_T1w.csv and csa-SC_T2w.csv result files, assuming they are located in the current folder.
+    csv_files = glob.glob('csa-SC_T*w.csv')
+
+    if not csv_files:
+        raise RuntimeError("Variable 'csv_files' is empty, i.e. no *.csv files were found in current directory.")
+
+    # Loop across *csv files and read them as a pandas DF
+    t1_t2_list = list()
+    for csv_file in csv_files:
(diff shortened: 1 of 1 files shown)


jcohenadad:
> **CONS** - with this workaround, subjects listed only in `csa_t1` (or only in `csa_t2`) are excluded not only from t1-t2 agreement figure, but also from `fig_csa_t1.png` and `fig_csa_t2.png` bars figures. It is not ideal, because when subject has only bad `fig_csa_t1.png`, it is not necessary to exclude him also from `fig_csa_t2.png`. But, to ensure this, it would require bigger code refactoring because subjects are excluded for each metric (t1, t2,...) separatelly and grouped by sites (so, sub_ID is lost and only lists of values for sites are further processed).

that's a big cons (and unnecessary). Why not simply implement the fix inside `generate_figure_t1_t2()`, the rest of the code should not be impacted

valosekj:
> that's a big cons (and unnecessary). Why not simply implement the fix inside `generate_figure_t1_t2()`, the rest of the code should not be impacted
> 
> e.g., in the loop below, just check if the subject exists for csa2, and if not, do not append:
> 
> https://github.com/spine-generic/spine-generic/blob/d05938623957d494cabb75bbb5331a572e06ba43/spinegeneric/cli/generate_figure.py#L541

This loop iterates across sites, not across subjects. At this step, values from individual subjects are grouped into list for each site and it is not apparent which specific subject was removed for given site.

(content of csa_t1 dict in this step is following)
image

Unfortunately, subjects are grouped by sites (i.e., the information which value is from which subject is removed) at early step of the script in the `aggregate_per_site` function:
https://github.com/spine-generic/spine-generic/blob/d05938623957d494cabb75bbb5331a572e06ba43/spinegeneric/cli/generate_figure.py#L219

I am continuing to think about appropriate workaround.

valosekj:
So, I have done little bit larger refactoring of `generate_figure_t1_t2` function in https://github.com/spine-generic/spine-generic/pull/199/commits/22c2088d24b3126ecdfe98c53206b3564f835dea because I did want to largely modify the whole script.
Now, the script excludes only bad `csa_t1` subjects for `fig_csa_t1.png `, only bad `csa_t2` subjects for `fig_csa_t2.png ` **and both bad `csa_t1` and `csa_t2` subjects for `fig_t1_t2_agreement_per_vendor.png`**
The refactored `generate_figure_t1_t2` function now fetch input `csa-SC_T1w.csv`, `csa-SC_T2w.csv` and `participants.tsv` and create pandas DF `t1_t2_df` as following:
image

Then, this is easy to exclude subjects and create figures.

TODO: update tests

jcohenadad:
@valosekj i find this refactoring quite overwhelming (was it necessary?), if you don't mind, i'll have a quick look at how we can do this in a simpler way

valosekj:
> @valosekj i find this refactoring quite overwhelming (was it necessary?), if you don't mind, i'll have a quick look at how we can do this in a simpler way

I know. It looks overwhelming. Some simpler solution is probably possible, but it would require changes at the beginning of the script (probably in `aggregate_per_site` function), which can then influence the rest of the script: 
 https://github.com/spine-generic/spine-generic/blob/4e04f54d8b9a77a0c97e3e77e8d34d82b5dbda74/spinegeneric/cli/generate_figure.py#L219

Because, in this function, individual subjects values are grouped by sites and information which value is from which subject is lost. It is okay for figure displaying only one metric. But, when we need to work with t1 and t2 and need to exclude subject from t2 based on t1, situation becomes complicated because only values grouped by sites are available.
The current proposed solution looks working (tests passed after last fix commit), so if you do not mind that script has some extra line, you do not have to invest your time into this.

jcohenadad:
I would really like to keep the code clean. Right now, there is a lot of duplication, e.g.:

re-reading the csv files (note: the code will break if not launched from curdir, which is _not_ a valid assumption because of the flag `-path-results`):
https://github.com/spine-generic/spine-generic/blob/4e04f54d8b9a77a0c97e3e77e8d34d82b5dbda74/spinegeneric/cli/generate_figure.py#L564

re-fetching values (already fetched before)
https://github.com/spine-generic/spine-generic/blob/4e04f54d8b9a77a0c97e3e77e8d34d82b5dbda74/spinegeneric/cli/generate_figure.py#L548

in general, this should be avoided because a modification of the structure (e.g. output of SCT's csv) will imply working at two different places in the code.

i'm now looking into a way to add an "exclude" vector for the df, which is input/output here:
https://github.com/spine-generic/spine-generic/blob/4e04f54d8b9a77a0c97e3e77e8d34d82b5dbda74/spinegeneric/cli/generate_figure.py#L308

since this is a new DF column, it will not interfere with the current codebase, and we can use it elegantly to exclude subjects (simple 'and' logical assertion-- should be one line of code)

jcohenadad:
Working on https://github.com/spine-generic/spine-generic/tree/jca/193-t1-t2-figure-fix,

The added field in `results_dict` now enables to cross-reference values with subjects:
````python
df['subject']['amu']
['sub-amu04', 'sub-amu05', 'sub-amu02', 'sub-amu01', 'sub-amu03']
df['val']['amu']
[0.0005520837445402861, 0.000467393490773328, 0.0005314071125280237, 0.00047408375044293656, 0.00042480097149554327]
````

jcohenadad:
@valosekj can you give this a try? https://github.com/spine-generic/spine-generic/pull/201

jcohenadad:
duplicate of https://github.com/spine-generic/spine-generic/pull/201

PR #201, opened 2020-09-08
Fix different x and y size in t1-t2 agreement figure

### LINKED ISSUE #193: generate_figure: ValueError: x and y must be the same size occurs if T1w and T2w subjects do not match

jcohenadad:
Fixes #193 by adding a vector of subject in the dataframe, enabling to check potential mismatch in `generate_figure_t1_t2`.

Also clarified code of `compute_regression()` by moving higher level stuff (dict, vendor info) to the caller.

Alternative approach to https://github.com/spine-generic/spine-generic/pull/199

### DIFF
--- a/spinegeneric/cli/generate_figure.py
+++ b/spinegeneric/cli/generate_figure.py
@@ -540,16 +540,22 @@ def generate_figure_t1_t2(df, csa_t1, csa_t2):
 
     # Create dictionary with CSA for T1w and T2w per vendors
     CSA_dict = defaultdict(list)
-    for index, line in enumerate(csa_t1):       # loop through individual sites
-        CSA_dict[line[1] + '_t1'].append(np.asarray(csa_t1[index, 3]))      # line[1] denotes vendor
-        CSA_dict[line[1] + '_t2'].append(np.asarray(csa_t2[index, 3]))      # line[1] denotes vendor
+    # loop across sites
+    for index, line in enumerate(csa_t1):
+        # Loop across subjects, making sure to only populate the dictionary with subjects existing both for T1 and T2
+        for subject in csa_t1[index][4]:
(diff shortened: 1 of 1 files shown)


valosekj (approved):
So smart 👍 Tested and works fine!

Issue #204, opened 2020-09-10
Automatize text generation based on results


jcohenadad:
When writing the results section of the spine-generic manuscript, there is a lot of manual copy/paste of the statistics, which is slow and prone to human error.

A better way would be to generate the text and auto-fill the results. Here would be the template:
~~~
The intra-site coefficients of variation (COVs) were averaged for each vendor and found to be all just under X%. The inter-site COVs (and inter-site ANOVA p-values) were X% (p=X) for GE, X% (p=X) for Philips and X% (p=X) for Siemens. The inter-vendor difference was significant (p=X), with the Tukey test showing significant differences between {SITE_X} and {SITE_Y} (p-adjusted Z) [and between …]
~~~

A few specifications for the implementation:
- This feature could be activated with the flag `-output-text`, which would write the text, for each contrast, into a text file (append to the same file). 
- The main would call a function `output_text()` that would take care of the formatting and writing into file
- P-values would need to be rounded (e.g 2 decimals?). Either we display `p<0.01` if smaller than that threshold, or we use scientific notation.

valosekj:
Are you working on this? Otherwise, I could implement it.

jcohenadad:
> Are you working on this? Otherwise, I could implement it.

that would be awesome!

jcohenadad:
p.s. i'm editing the description of the issue to add implementation specifications-- please feel free to add specs and/or comment on them. Thanks!

PR #207, opened 2020-09-12
Automatize text generation based on results

valosekj:
This PR automatizes text generation based on results.
The feature is activated with the flag `-output-text`, then the `main` calls a function `output_text()` which performs writing of results into `statistical_results.txt` file.
P-values are rounded to two decimals, if p<0.01 (e.g. 0.003), it is replace by str "<0.01".

The output `statistical_results.txt` looks like:
Snímek obrazovky 2020-09-12 v 20 46 45

Fixes: https://github.com/spine-generic/spine-generic/issues/204

### DIFF
--- a/spinegeneric/cli/generate_figure.py
+++ b/spinegeneric/cli/generate_figure.py
@@ -19,6 +19,7 @@ import pandas as pd
 import subprocess
 from textwrap import dedent
 import yaml
+import math
 
 import numpy as np
 from scipy import ndimage
@@ -190,6 +191,11 @@ def get_parser():
         action='store_true',
         help="Show 95%% confidence interval on the plot.")
     parser.add_argument(
+        '-output-text',
+        required=False,
+        action='store_true',
+        help="Write statistics results into text file.")
+    parser.add_argument(
         '-exclude',
         required=False,
         help=
@@ -356,8 +362,8 @@ def compute_statistics(df):
         # ANOVA: category=[site]
         values_per_site = [df['val'][(df['vendor'] == vendor) & (df['site'] == i_site)][0]
                            for i_site in df['site'][df['vendor'] == vendor]]
-        stats['anova_site'][site] = f_oneway(*values_per_site)
-        logger.info("ANOVA[site] for {}: {}".format(vendor, stats['anova_site'][site]))
+        stats['anova_site'][vendor] = f_oneway(*values_per_site)
+        logger.info("ANOVA[site] for {}: {}".format(vendor, stats['anova_site'][vendor]))
 
     # ANOVA: category=[vendor]
     stats['anova_vendor'] = f_oneway(*[df['mean'][df['vendor'] == i_vendor] for i_vendor in vendors])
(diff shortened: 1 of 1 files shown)


spinegeneric/cli/generate_figure.py
```diff
@@ -368,6 +374,88 @@ def compute_statistics(df):
...
+        elif count == 1:
+            file.write(" and ")
+        elif count == 2:
+            file.write(". ")
+
+    p_val_anova = stats['anova_vendor'][1]
+    # Write post-hoc Tukey results if inter-vendor difference was significant
+    if p_val_anova < 0.05:
+        p_val_anova = check_p_value(p_val_anova)
+        file.write("The inter-vendor difference was significant (p={}), with the Tukey test showing significant "
```
jcohenadad, line 428: with that implementation, the output text is `=<0.01` instead of `<0.01`. A suggestion would be to encode the "=" in the `check_p_value()`

spinegeneric/cli/generate_figure.py
```diff
@@ -368,6 +374,88 @@ def compute_statistics(df):
...
+        Check if p-value is lower than 0.01, if so, change it to "<0.01", otherwise, round it to two decimals
+        :param p_val: input p-value
+        :return: p_val: processed p-value (replaced by "<0.01" or rounded to two decimals)
+        """
+        if p_val < 0.01:
+            p_val = "<0.01"
+        else:
+            p_val = round(p_val, 2)
+
+        return p_val
```
jcohenadad, line 395: it is usually cleaner to stick with a single output format. Here, it is either a string or a float.

spinegeneric/cli/generate_figure.py
```diff
@@ -368,6 +374,88 @@ def compute_statistics(df):
...
 
 
+def output_text(stats, metric):
+    """
+    # Write statistical results into text file
+    :param stats: dict with stat resutls
+    :param metric: currently processed metric (e.g, dti_fa,...)
+    """
+
+    def check_p_value(p_val):
```
jcohenadad, line 384: the main purpose of this function is to format the output p-value (checking it is the "how"). Usually we name functions as "what" instead of "how", so i suggest naming it `format_p_value()`

spinegeneric/cli/generate_figure.py
```diff
@@ -368,6 +374,88 @@ def compute_statistics(df):
...
+    if not os.path.isfile(fname):
+        file = open(fname, "w+")
+    else:
+        file = open(fname, "a+")
+
+    # Write metric name and two blank lines
+    file.write('{}:\n\n'.format(metric))
+    # Find and write highest intra-site COV (rounded up)
+    file.write("The intra-site coefficients of variation (COVs) were averaged for each vendor and found to be all "
+                "just under {}%. ".format(math.ceil(max(stats['cov_intra'].values()) * 100)))
```
jcohenadad, line 408:
```suggestion
                "under {}%. ".format(math.ceil(max(stats['cov_intra'].values()) * 100)))
```

jcohenadad:
@valosekj i only left minor comments-- it looks great otherwise! very useful!

valosekj:
> @valosekj i only left minor comments-- it looks great otherwise! very useful!

Thank you! I'll implement corrections tomorrow.

spinegeneric/cli/generate_figure.py
valosekj: I am not sure if I get it. When p-val is lower than 0.01 (e.g. 0.009), it is changed to string <0.01. If the p-val is exactly equal to 0.01, p-val falls under _else_ statement and function returns back 0.01.

image

jcohenadad: line 428, the “=“ is hardcoded and is always displayed, while it should not if p<0.01

spinegeneric/cli/generate_figure.py
```diff
@@ -368,6 +374,88 @@ def compute_statistics(df):
     return df, stats
 
 
+def output_text(stats, metric):
```
valosekj, line 377: Don't we want to include example of full text somewhere inside script (docstring or help)? Or is not it necessary?

`The intra-site coefficients of variation (COVs) were averaged for each vendor and found to be all just under X%. The inter-site COVs (and inter-site ANOVA p-values) were X% (p=X) for GE, X% (p=X) for Philips and X% (p=X) for Siemens. The inter-vendor difference was significant (p=X), with the Tukey test showing significant differences between VENDOR_X and VENDOR_Y (p-adj=X) and between VENDOR_Y and VENDOR_Z (p-adj=X).`

spinegeneric/cli/generate_figure.py
valosekj: Thanks for explanation! Now, I get it how you meant it. Should by fixed in https://github.com/spine-generic/spine-generic/pull/207/commits/01f8108c98b5f57b67f5b1c7f64337e1b83a5bc0

spinegeneric/cli/generate_figure.py
```diff
@@ -368,6 +374,88 @@ def compute_statistics(df):
...
+    """
+    # Write statistical results into text file
```
jcohenadad, line 379:
```suggestion
    # Embed statistical results into sentences so they can easily be copy/pasted into a manuscript. 
```

spinegeneric/cli/generate_figure.py
jcohenadad: no need for example-- i've updated the docstrings to clarify what this function does

valosekj:
> @valosekj i only left minor comments-- it looks great otherwise! very useful!

I have implemented all suggestions --> should be ready for review

jcohenadad (approved):
awesome 👍

PR #215, opened 2020-09-15
Cosmetic improvements with formatted output of statistics

### LINKED ISSUE #214: Small issues with statistical results
Few issues noticed with commit 18699b647ad2fdd24dee74ada6b42b474814498a: 

- When using the [single-subject dataset r20200914](https://github.com/spine-generic/data-single-subject/releases/tag/r20200914), "p=nan" values are found (e.g. below for the CSA_t1):
  ~~~
  The intra-site coefficients of variation (COVs) were averaged for each vendor and found to be all under 0%. The inter-site COVs (and inter-site ANOVA p-values) were 2.3% (p=nan) for GE, 1.8% (p=nan) for Philips and 0.9% (p=nan) for Siemens. The inter-vendor difference was significant (p<0.01), with the Tukey test showing significant differences between GE and Philips (p-adj=0.03) and between GE and Siemens (p-adj<0.01).
  ~~~
  - The reason is that we cannot compute an ANOVA with singleton per site. So, results should be skipped in case of single subject.
- Also, we don't need to re-define COVs every time. So we should instead write: "The intra-site COVs were averaged..."
- The file `log_stats.txt` should be deleted (if it already exists), otherwise values are appended in case we run the stats multiple times.
- If possible (cherry on the cake), we should remove the blurb about the intra-site COV if it is equal to 0 (ie: single-subject case).

jcohenadad:
This PR brings few improvements for the formatting of statistics output:
- Skip intra-site ANOVAs and pvalues if single-subject
- Remove the blurb about the intra-site COV if it is equal to 0 (ie: single-subject case).
- Removed COVs acronym definition
- Remove existing output txt file
- Output formatted statistics into the same log_stats.txt file

Fixes #214

### DIFF
--- a/spinegeneric/cli/generate_figure.py
+++ b/spinegeneric/cli/generate_figure.py
@@ -396,18 +396,31 @@ def output_text(stats):
 
     txt = ""
 
-    # Write metric name and two blank lines
+    # Find out if this is single-subject data (single value per site)
+    if max(stats['cov_intra'].values()) == 0.0:
+        single_subject = True
+    else:
+        single_subject = False
+
     # Find and write highest intra-site COV (rounded up)
-    txt += "The intra-site coefficients of variation (COVs) were averaged for each vendor and found to be all under " \
-           "{}%. ".format(math.ceil(max(stats['cov_intra'].values()) * 100))
+    if not single_subject:
+        txt += "The intra-site coefficients of variation (COVs) were averaged for each vendor and found to be all under " \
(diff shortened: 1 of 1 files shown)


valosekj (approved):
Tested on both single- and multi-subject datasets and works fine 👍
Only, is this rounding to `=0.05` optimal? Maybe `p-adj<0.05` would be better in this case.

Snímek obrazovky 2020-09-15 v 19 37 16

jcohenadad:
> Only, is this rounding to =0.05 optimal? Maybe p-adj<0.05 would be better in this case.

it's fine

PR #226, opened 2020-09-25
Compute min, max and median age across subjects

valosekj:
This PR adds computation of min, max and median age across subjects from `participants.tsv` file
These stats are computed across all subjects (i.e., no exclusion is used).
Stats are then saved in form of prepared sentence for manuscript into `log_stats.txt` file:

`..., age between 19.0 and 56.0 y.o., median age 28.0 y.o..`

~~Known disadvantage of proposed solution - participants file is read for every metric in `aggregate_per_site` function:~~
 https://github.com/spine-generic/spine-generic/blob/de8b6c9df0e4a28ad418be7f990cf726cb7e6bfd/spinegeneric/cli/generate_figure.py#L234-L235
~~Then, age stats are also saved into `log_stats.txt` file for every metric separately. I could add some counter to save age only one time. Should I?~~

UPDATE: separate function was created https://github.com/spine-generic/spine-generic/pull/226/commits/d3975dfd9ed037f3ab8535bbb488a233a5645678, printing age stats only once

### DIFF
--- a/spinegeneric/cli/generate_figure.py
+++ b/spinegeneric/cli/generate_figure.py
@@ -234,6 +234,12 @@ def aggregate_per_site(dict_results, metric, dict_exclude_subj):
     # Build Panda DF of participants based on participants.tsv file
     participants = pd.read_csv(os.path.join('participants.tsv'), sep="\t")
 
+    # Compute min, max and median for age across all subjects and save it to log
+    age_stat = participants['age'].agg(['median', 'min', 'max'])
+    logger.info('..., age between {} and {} y.o., median age {} y.o..'.format(age_stat['min'],
+                                                                              age_stat['max'],
+                                                                              age_stat['median']))
+
     # Fetch specific field for the selected metric
     metric_field = metric_to_field[metric]
     # Build a dictionary that aggregates values per site
--- a/spinegeneric/cli/generate_figure.py
+++ b/spinegeneric/cli/generate_figure.py
@@ -223,7 +223,7 @@ def get_parser():
     return parser
 
 
-def aggregate_per_site(dict_results, metric, dict_exclude_subj):
+def aggregate_per_site(dict_results, metric, dict_exclude_subj, counter):
     """
     Aggregate metrics per site. This function assumes that the file participants.tsv is present in the -path-results
     folder.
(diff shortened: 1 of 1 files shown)


jcohenadad:
> Then, age stats are also saved into log_stats.txt file for every metric separately. I could add some counter to save age only one time. Should I?

yes, sounds good to me

valosekj:
> > Then, age stats are also saved into log_stats.txt file for every metric separately. I could add some counter to save age only one time. Should I?
> 
> yes, sounds good to me

Should by done in https://github.com/spine-generic/spine-generic/pull/226/commits/5ee7ca1e025c73df8e8cb52697f0c2cc71312c48

jcohenadad (changes requested):
@valosekj sorry for the nitpicking, but after giving it some thoughts, it would be much cleaner to simply create a new function that loads the participants.tsv and output the age stats, and the function would be run only once. Because this operation is outside the scope of "aggregating metrics". You could call it eg: `compute_age_statistics()`.

it's important to keep the code as clean as possible because others will look at it, and might be use it for their own research.

valosekj:
Makes sense! I will do in a few hours.

valosekj:
I have created separate function for loading of `participants.tsv` file and another function for age stats computation in last commit https://github.com/spine-generic/spine-generic/pull/226/commits/d3975dfd9ed037f3ab8535bbb488a233a5645678. 
Age stats are computed only once and printed at the beginning of `log_stats.txt` file. So far like this:
```
Age statistics:
..., age between 19.0 and 56.0 y.o., median age 28.0 y.o..
```

Or should I include whole sentence from manuscript (it is little bit long).
```
In order to evaluate a more realistic (routine) scenario, different groups worldwide with varying levels of expertise each
scanned six different healthy subjects (3 males, 3 females), aged between 19 and 56 y.o., median age 28 y.o..
```

I also realized that now, age stats are computed for both single- and multi- subject datasets. For single subject, it does not make so much sense, but it is not wrong. Do we want deal with this now and print age stats selectively only for multi-subject dataset?

jcohenadad:
> Or should I include whole sentence from manuscript (it is little bit long).

too long-- unecessary-- we just need the numbers

> I also realized that now, age stats are computed for both single- and multi- subject datasets. For single subject, it does not make so much sense, but it is not wrong. Do we want deal with this now and print age stats selectively only for multi-subject dataset?

don't worry about it

valosekj:
Okay, then it should be ready to review.

jcohenadad (approved):
nice!

Issue #90, opened 2020-06-23
date_of_scan info missing in participants.tsv


jcohenadad:
concerns the single-subject dataset https://openneuro.org/datasets/ds002900

valosekj:
It looks that only `sub-perform ` contains info about date of scan (`AcquisitionDateTime` tag) in json sidecars. Remaining subjects contain only `AcquisitionTime` tag.

jcohenadad:
ah that's too bad 😞 . Thanks for looking into it @valosekj

jcohenadad:
fixed in https://openneuro.org/datasets/ds002900/versions/1.0.3

PR #249, opened 2021-05-25
Add preprocessing and keep segmentation only

sandrinebedard:
## Context
Related with [issue #90](https://github.com/spine-generic/data-multi-subject/issues/90).
The purpose of this PR is to generate a processed dataset of [spine-generic/data-multi-subject](https://github.com/spine-generic/data-multi-subject) using `preprocess_data.sh` and generate SC and GM segmentations with `process_data.sh`.
## Description
This PR includes:
* `process_data.sh` is seperated in 2 scripts --> `preprocess_data.sh` and `process_data.sh`
* Only segmentation is done in `process_data.sh`, all other processing steps are removed.
* Segmentation is also done for T2* and MTS data.
* `curate_derivatives_spine_generic.py` : curates the derivatives from spine-generic/data-multi-subject

## `preprocess_data.sh`
Preprocessing steps are:
For T1w and T2w :
* Resampling
* Reorient to RPI
* Remove suffix _RPI_r

For T2star:
* Compute root-mean square across 4th dimension (if it exists)
* Remove suffix _rms

For DWI:
* Generate mean image after motion correction
* Remove intermidiate files

### Questions:
* Do we need preprocessing on MTS data?
* Since SC segmentation was computed on mean image after motion correction for dwi, I kept it that way, and the mean image would be in the `spine-generic-processed dataset`, is that ok?

## `process_data.sh`
SC segmentation is performed on T1, T2. T2*, T1w_MTS, MTon_MTS and dwi
### Questions:
* Automatic segmentation fails on MToff_MTS data, I also tried with sct_propseg and it works better, but would require a lot of manual corrections. I would remove it.
* Do we keep T1w_MTS since we already have T1w contrast? (I would keep it, better to have more segmentations)

##  `curate_derivatives_spine_generic.py`
* Gets derivatives with suffix `_seg-manual`
* Removes suffix `_RPI_r` or `_rms`
* Creates a json sidecar if it does not exist
* Reorient and resamples `_csgseg-manual.nii.gz` images

### DIFF
--- a/process_data.sh
+++ b/process_data.sh
@@ -124,14 +124,8 @@ segment_if_does_not_exist $file_t2 "t2"
 file_t1w="${SUBJECT}_acq-T1w_MTS"
 file_mton="${SUBJECT}_acq-MTon_MTS"
 
-if [[ -e "${file_t1w}.nii.gz" && -e "${file_mton}.nii.gz" && -e "${file_mtoff}.nii.gz" ]]; then 
-  # Segment spinal cord (only if it does not exist)
-  segment_if_does_not_exist $file_t1w "t1" 
-  segment_if_does_not_exist $file_mton "t2s"
- 
-else
-  echo "WARNING: MTS dataset is incomplete."
-fi
+segment_if_does_not_exist $file_t1w "t1" 
+segment_if_does_not_exist $file_mton "t2s"
 
 
 # T2s
--- a/process_data.sh
+++ b/process_data.sh
@@ -148,18 +148,8 @@ segment_gm_if_does_not_exist $file_t2s "t2s"
 
 # DWI
 # ------------------------------------------------------------------------------
-file_dwi="${SUBJECT}_dwi"
 cd ../dwi
-if [[ -e ${SUBJECT}_acq-b0_dwi.nii.gz ]]; then
-  echo "Found additional b=0 scans: $file_b0.nii.gz"
-  # Update global variable
-  FILE_DWI="${file_dwi}_concat"
-else
-  echo "No additional b=0 scans was found."
-  FILE_DWI="${file_dwi}"
-fi
-
-file_dwi_mean="${FILE_DWI}_moco_dwi_mean" # TO RENAME
+file_dwi_mean="${SUBJECT}_rec-average_dwi"
 
 # Segment spinal cord (only if it does not exist)
 segment_if_does_not_exist ${file_dwi_mean} "dwi"
@@ -176,7 +166,7 @@ FILES_TO_CHECK=(
   "anat/${SUBJECT}_acq-MTon_MTS_seg.nii.gz"
   "anat/${SUBJECT}_T2star_seg.nii.gz"
   "anat/${SUBJECT}_T2star_gmseg.nii.gz"
-  "dwi/${FILE_DWI}_moco_dwi_mean_seg.nii.gz" # MAYBE change if we rename
+  "dwi${SUBJECT}_rec-average_dwi_seg.nii.gz"
 )
 for file in ${FILES_TO_CHECK[@]}; do
   if [[ ! -e $file ]]; then
--- a/process_data.sh
+++ b/process_data.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Process data.
+# Process data, modfified version to only create SC segmentations and GM segmentation for T2s. Uses spine-generic-processed as an input.
 #
 # Usage:
 #   ./process_data.sh <SUBJECT>
@@ -8,7 +8,7 @@
 # Manual segmentations or labels should be located under:
 # PATH_DATA/derivatives/labels/SUBJECT/<CONTRAST>/
 #
-# Authors: Julien Cohen-Adad
+# Authors: Julien Cohen-Adad, Sandrine Bédard
 
 # The following global variables are retrieved from the caller sct_run_batch
 # but could be overwritten by uncommenting the lines below:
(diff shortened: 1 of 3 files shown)


jcohenadad (commented):
looks good! is there any reason for splitting preprocessing and processing? keeping only one script is simpler in that we can point to a single input directory (whereas with 2 scripts, the "process_data.sh" script uses as input the output of preprocess_data.sh, which makes it slightly more difficult to track-- although not impossible...).

also, we need to decide wether we want to keep those modified scripts there, as it is project-specific. Maybe we should put them (or it, if we keep only one script) under https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord, and name it something like: `process_data_spinegeneric.sh` (and we would have other scripts like `process_data_ukbiobank.sh`, etc. that would be specific to the SC segmentation project).

sandrinebedard:
> Is there any reason for splitting preprocessing and processing? 

The output of preprocess_data.sh would directly be the dataset `spine-generic-processed`, we want those images as an input for training, that's why I seperated them 🙂

> also, we need to decide wether we want to keep those modified scripts there, as it is project-specific

Yes we could move the scripts there I agree! Well it is project specific, but it only generates segmentations, so not that specific

jcohenadad:
> > Is there any reason for splitting preprocessing and processing?
> 
> The output of preprocess_data.sh would directly be the dataset `spine-generic-processed`, we want those images as an input for training, that's why I seperated them 🙂

ah! of course 🤦  that makes perfect sense

> > also, we need to decide wether we want to keep those modified scripts there, as it is project-specific
> 
> Yes we could move the scripts there I agree! Well it is project specific, but it only generates segmentations, so not that specific

indeed. Especially for the "processing" script which should be dataset agnostic (as long as the dataset is BIDS-friendly)

sandrinebedard:
I woud be ready to start the preprocessing if the script is all good in that case!

process_data.sh
```diff
@@ -146,162 +108,77 @@ cd ${SUBJECT}/anat/
...
-  "anat/${SUBJECT}_T2star_rms_gmseg.nii.gz"
-  "dwi/dti_FA.nii.gz"
-  "dwi/dti_MD.nii.gz"
-  "dwi/dti_RD.nii.gz"
-  "dwi/label/atlas/PAM50_atlas_00.nii.gz"
+  "anat/${SUBJECT}_T1w_seg.nii.gz"
+  "anat/${SUBJECT}_T2w_seg.nii.gz"
+  "anat/${SUBJECT}_acq-T1w_MTS_seg.nii.gz" # Maybe remove
+  "anat/${SUBJECT}_acq-MTon_MTS_seg.nii.gz"
+  "anat/${SUBJECT}_acq-MToff_MTS_seg.nii.gz" # Maybe remove --> automatic seg fails
```
jcohenadad, line 178: I agree, you can remove

process_data.sh
```diff
@@ -146,162 +108,77 @@ cd ${SUBJECT}/anat/
...
+if [[ -e ${SUBJECT}_acq-b0_dwi.nii.gz ]]; then
+  echo "Found additional b=0 scans: $file_b0.nii.gz"
+  # Update global variable
+  FILE_DWI="${file_dwi}_concat"
+else
+  echo "No additional b=0 scans was found."
+  FILE_DWI="${file_dwi}"
+fi
+
+file_dwi_mean="${FILE_DWI}_moco_dwi_mean"
```
jcohenadad, line 163: we probably need to find a BIDS-compliant name for it... @alexfoias might have a clue

sandrinebedard: I searched a bit, but I did find anything really relevant, right now, the name is for example:
`sub-amu01_dwi_concat_moco_dwi_mean.nii.gz` (I definitely would not keep that)
I found this --> https://bids-specification.readthedocs.io/en/stable/05-derivatives/02-common-data-types.html#preprocessed-or-cleaned-data
So we should add a json sidecar I think
Maybe the name could be : `sub-amu01_dwi_average.nii.gz`

mariehbourget: `sub-amu01_dwi_average.nii.gz` is definitely not compliant.
As for `sub-amu01_desc-average_dwi.nii.gz`, it is compliant, but `desc-` is an entity only allowed in the `derivatives` folder (see: https://bids-specification.readthedocs.io/en/stable/99-appendices/09-entities.html#desc)

So it depends on what you are doing here. Are those derivatives in an existing dataset or is this a new separate "preprocess" dataset?

We had a similar discussion thread about the vertebral labelling script [here](https://github.com/ivadomed/ivadomed/issues/707#issuecomment-829209716), where we decided to use the entity `rec-`. Maybe it could apply here.
`rec-` definition: https://bids-specification.readthedocs.io/en/stable/99-appendices/09-entities.html#rec

sandrinebedard: > Are those derivatives in an existing dataset or is this a new separate "preprocess" dataset?

It's going to be a new preprocess dataset, so yes `rec-` would be good I think here! Thanks @mariehbourget 
So it would be something like : `sub-amu01_rec-average_dwi.nii.gz` ?
motion correction was also applied before averaging, we should add it?

jcohenadad: > motion correction was also applied before averaging, we should add it?

no need to elongate the file name further-- that information should be present in the README of the repository, which will/should include a link to the processing file (used to generate this processed dataset).

thank you @mariehbourget 🙏

sandrinebedard:
I ran preprocessing on all the data of `data-multi-subject`
Preprocessing for dwi failed for:
* `sub-oxfordOhba02`
* `sub-ucdavis04`
* `sub-ucdavis01`
* `sub-ucdavis02`

For `sub-ucdavis04`, `sub-ucdavis01`, `sub-ucdavis02`, we have 3 scans for dwi :

* `sub-ucdavis01_run-01_dwi.nii.gz`
* ` sub-ucdavis01_run-02_dwi.nii.gz`
* `sub-ucdavis01_run-03_dwi.nii.gz`

That's why it was not taken into account. I can modify the script to inlcude it, in that case which one the scan do I chose?

We can also ignore that and have a missing modality for those 3 subjects.

jcohenadad:
> We can also ignore that and have a missing modality for those 3 subjects

we have enough data for this project, you can ignore

sandrinebedard:
@jcohenadad I finished the script to curate the derivatives from `spine-generic/data-multi-subject` for `spine-generic-processed` (`curate_derivatives_spine_generic.py`) if you could check it out quickly before I add the curated derivatives to the dataset `spine-generic-processed`, thanks 🙂

jcohenadad (approved):
I haven't tested but LGTM!

jcohenadad:
@sandrinebedard could you please move these scripts into the [contrast-agnostic repos](https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord) (and update the doc accordingly)

sandrinebedard:
Closing this PR since the script and the information has been moved to https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord#readme.

PR #260, opened 2022-09-26
Update skimage to the latest

kousu:
skimage 0.17 does not have wheels built for python 3.9 or python 3.10 because they didn't exist when it was published, meaning anyone who tries to use spine-generic on a modern system (even just Ubuntu 22.04 LTS) will find themselves in a surprise compilation mire.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ sklearn
 statsmodels
 pyyaml
 pybids~=0.13.0 # pinned until https://github.com/bids-standard/pybids/pull/791 is fixed
-scikit-image~=0.17.2
+scikit-image
 transforms3d~=0.3.1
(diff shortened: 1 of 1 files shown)


jcohenadad:
good catch!

kousu:
Originally I was using this for the demo version of https://spineimage.ca, which uses python 3.10, but now @jcohenadad I've expanded this into a bit of an omnibus maintenance PR. It shouldn't be anything wild.

If you're too busy to look it over is there anyone else who knows this code? Or if you feel confident in it, maybe you can just approve it without having to look too closely.

jcohenadad (approved):
I have not tested the code, but reading through it I understand it and the proposed changes make sense. Thank you for looking into it @kousu

PR #262, opened 2022-09-29
Move flake8 configuration into .flake8

kousu:
It is easier (and more conventional) to maintain.

Also, make lint.yml treat all flake warnings as errors, not just syntax problems. Previously, the linter had been turned down so that only four errors were actually errors, and the rest got --exit-zero, presumably to triage problems, but the time I found it both had been entirely disabled. I separated linting to a separate Action which I made non-mandatory, achieving the same effect, so now I think we can shorten the script to just run the full linter in the conventional way.

And if we can get the lint back under control then we can make it mandatory again.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,10 @@
+[flake8]
+
+# This constrains how dense the code of each function can be
+max-complexity = 10
+
+# The GitHub editor is 127 chars wide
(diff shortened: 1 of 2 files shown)


jcohenadad (approved):
looks good

PR #261, opened 2022-09-27
Reformat code via `black`.

kousu:
There's a lot of edits but all were produced by a robot:

```
(py39) p115628@joplin:~/src/spine-generic$ black .
reformatted spinegeneric/__init__.py
reformatted spinegeneric/bids.py
reformatted spinegeneric/tests/test_copy_to_derivatives.py
reformatted spinegeneric/tests/test_cli_manual_correction.py
reformatted spinegeneric/tests/test_populate_derivatives.py
reformatted spinegeneric/tests/test_generate_figure.py
reformatted setup.py
reformatted spinegeneric/cli/copy_to_derivatives.py
reformatted spinegeneric/cli/populate_derivatives.py
reformatted docs/conf.py
reformatted spinegeneric/cli/qc_bids_deface.py
reformatted spinegeneric/cli/deface_spineGeneric_usingR.py
reformatted spinegeneric/cli/check_data_consistency.py
reformatted spinegeneric/cli/package_for_correction.py
reformatted spinegeneric/cli/params_checker.py
reformatted spinegeneric/cli/manual_correction.py
reformatted spinegeneric/utils.py
reformatted spinegeneric/cli/create_mosaic.py
reformatted spinegeneric/cli/generate_figure.py

All done! ✨ 🍰 ✨
19 files reformatted, 4 files left unchanged.
```

Follows #260.

PR #263, opened 2022-09-29
Fix most flake8 errors.

kousu:
This follows #261 ~~(hence is Drafted until that is merged)~~. That one was fully mechanical, but this one I did by hand. It needs a closer inspection. I don't think I changed the meaning of any code, but I might have made a typo or not understood some outer loop.

Some of the errors, particularly the "--max-complexity=10" and "module level import not at top of file", require actual refactoring work, so I am not touching them here.

Compared to #261, this has reduced the linting errors by 5x:

```
p115628@joplin:~/src/spine-generic$ flake8 . --count --quiet --max-line-length=127 --max-complexity=10
./spinegeneric/utils.py
./spinegeneric/cli/deface_spineGeneric_usingR.py
./spinegeneric/cli/__init__.py
./spinegeneric/cli/params_checker.py
./spinegeneric/cli/populate_derivatives.py
./spinegeneric/cli/generate_figure.py
./spinegeneric/cli/check_data_consistency.py
./spinegeneric/cli/qc_bids_deface.py
./spinegeneric/cli/manual_correction.py
./spinegeneric/cli/create_mosaic.py
./spinegeneric/cli/package_for_correction.py
./docs/conf.py
57
p115628@joplin:~/src/spine-generic$ git checkout ng/flakes
Basculement sur la branche 'ng/flakes'
p115628@joplin:~/src/spine-generic$ flake8 . --count --quiet --max-line-length=127 --max-complexity=10
./spinegeneric/cli/params_checker.py
./spinegeneric/cli/generate_figure.py
./spinegeneric/cli/manual_correction.py
./spinegeneric/cli/create_mosaic.py
./spinegeneric/cli/package_for_correction.py
./docs/conf.py
12
```


jcohenadad (approved):
thank you

spinegeneric/cli/generate_figure.py
```diff
@@ -527,7 +526,7 @@ def format_p_value(p_val):
 
         # Get significant post-hoc results
         index = sum(
-            stats["tukey_test"].reject == True
+            bool(stats["tukey_test"].reject)
```
kousu, line 529: I have second-thoughts about this line. `flake8` flags any `== True` as a mistake, but this is doing a vectorized comparison. In fact you shouldn't be able to call `sum(bool(_))` because the types are mismatched: `sum()` takes a list.

I would have to take some time to learn how to/write a test for this line before this can be merged.

spinegeneric/cli/generate_figure.py
```diff
@@ -1317,7 +1312,7 @@ def main(argv=sys.argv[1:]):
 
         # Excluded sites
         logger.info(
-            "Sites removed: {}".format(list(df[df["exclude"] == True]["site"].values))
+            "Sites removed: {}".format(list(df[df["exclude"]]["site"].values))
```
kousu, line 1315: Ditto: this was a vectorized operation. Is it still?

spinegeneric/cli/generate_figure.py
mguaypaq: I think pandas (and numpy) get a special pass for the flake8 lints about comparison operators, since they have unexpected (to flake8) semantics. I reinstated the `== True` comparison and added a `noqa: 712` comment to silence the warning.

spinegeneric/cli/generate_figure.py
mguaypaq: Ditto: reinstated the `== True` comparison and added a `noqa: 712` comment.

mguaypaq (approved):
I fixed the merge conflicts with `master` by taking their version of `params_checker.py`, because it's been changed in #265.

I disabled the max-complexity check, because I don't think that's actionable advice, and fixed the few remaining flake8 linting complaints. I also reinstated the `== True` comparisons for pandas DataFrames, because those behave specially, and shouldn't be subject to flake8's complaints.

Having quickly looked at the other changes, they seem fine (and the `importlib.metadata` trick is very neat!). I'm going to merge this because I'm tired of getting emails about failed linting checks in all the other PRs :wink:

PR #270, opened 2023-08-29
Migrate ReadTheDocs configuration to `.readthedocs.yaml`

### LINKED ISSUE #268: ReadTheDocs builds failing due to missing `results.zip` asset on recent `data-multi-subject` releases
@valosekj reported on Slack that the documentation seems to no longer be updating:
> Hey! Do you know why https://spine-generic.readthedocs.io/ still shows the old content, although [my PR](https://github.com/spine-generic/spine-generic/pull/267) was merged several days ago?

Looking at [the builds page](https://readthedocs.org/projects/spine-generic/builds/), it looks like the ReadTheDocs builds have been failing since [November 15, 2021](https://readthedocs.org/projects/spine-generic/builds/15274509/), and the last successful build was [October 6, 2021](https://readthedocs.org/projects/spine-generic/builds/14915013/) (almost 2 years ago).

This is maybe also the reason for #259. Maybe also related to #250, although that was before, and wasn't reproducible.

mguaypaq:
Fixes #268. This migration has to happen before September 25, according to [this ReadTheDocs blog post](https://blog.readthedocs.com/migrate-configuration-v2/).

~~This PR is currently a draft, so that I can see how the settings affect an actual build.~~

This PR also includes a few fixes to:
- Run the documentation build on Python 3.11, which is the [current default version supported by ReadTheDocs](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python).
- Address some formatting warnings in the documentation build.

The [most recent build](https://readthedocs.org/projects/spine-generic/builds/21902243/) for commit c5e70252caca6a9864b86084a090e2d3c081bc60 succeeded.


mguaypaq:
Hmm, this PR doesn't appear in the [list of potential builds](https://readthedocs.org/projects/spine-generic/versions/?version_filter=mgp) on ReadTheDocs, maybe because it's a draft? I'll un-draft it.

docs/index.rst
```diff
@@ -29,7 +29,7 @@ Contributors
 ------------
 
 A list of contributors for the analysis pipeline is available `here <https://github.com/spine-generic/spine-generic/graphs/contributors>`__.
-If you would like to contribute to this project, please see `contribution guidelines <https://github.com/spine-generic/spine-generic/blob/master/CONTRIBUTING.md>`_.
```
joshuacwnewton, line 32: I'm curious about the motivation for each of these style changes. I presumed that they would be necessary to placate Sphinx's fail_on_warning setting, yet it seems to be absent from the new  `.readthedocs.yaml` file?

Is that something we want to add, as was done [for SCT](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/45412456dfb7a54542e1eb8aa0eb6ca8ffbb44db/.readthedocs.yaml#L25-L28)? Or are there still too many warnings to want to turn that on?

joshuacwnewton (approved):
The changes as-is LGTM! I just had a few comments on things that could potentially be added to this PR (or a follow-up PR, honestly).

spinegeneric/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "2.5dev"
```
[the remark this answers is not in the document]
mguaypaq: Good point, I hadn't noticed this. It looks like the conversion `X.Ydev -> X.Y.dev0` is a feature of setuptools, according to my reading of [their documentation on specifying versions](https://setuptools.pypa.io/en/latest/userguide/distribution.html), and I guess the normalized version number gets written in `${PKG}.egg-info/PKG-INFO`.

setup.py
```diff
@@ -31,6 +31,7 @@
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
```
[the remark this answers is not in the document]
mguaypaq: Good point, done in c0e405aff35b705884bbdf59f182105738fc44ad.

docs/index.rst
mguaypaq: They were just warnings from my local Sphinx build, or inspired from what I learned by fixing warnings, not from any automated checks. A short explanation:
- *For changing links from `_` (single underscore) to `__` (double underscore):* Apparently using a single underscore at the end of a hyperlink, like
  <pre><code>`my link text &lt;https://example.com/my/link/url.html>`_</code></pre>
  creates a document-wide link target, keyed by `my link text` (see [here](https://docutils.sourceforge.io/docs/user/rst/quickref.html#hyperlink-targets), maybe? The documentation is not great). And Sphinx doesn't want multiple targets with the same text, which generated a warning. To avoid this, the advice is to use double underscore, which I tried to apply uniformly to all links.
- *For the whitespace in a table row:* There was a mismatch with the surrounding lines, and Sphinx actually complained about it, so I fixed it. Possibly the result of a search-and-replace that didn't take into account word lengths.
- *For removing some of the header link targets:* Most of them were unused? Maybe I could have left them in. But I think the rule should probably be either "Only used targets get defined" or "All headers get targets", and I didn't feel like adding all those targets. This was triggered because one of the used targets wasn't defined, causing a warning (and a dead link).

mguaypaq: > Ah, I just realised that SG doesn't have RTD building set up for PRs, so the utility of fail_on_warning (CI checks) is lost.

Yeah, actually setting that up seemed out of scope.

PR #279, opened 2024-10-09
Specify sphinx-rtd-theme as a dependency

mguaypaq:
The ReadTheDocs page has been failing to build for the past year:

https://readthedocs.org/projects/spine-generic/builds/

Looking at the logs, it looks like this is because of a missing dependency in `requirements.txt` for the ReadTheDocs theme. According to the theme documentation for the latest version (3.0.1), it looks like the theme needs to be pip installed, and activated as an extension, and they recommend pinning the version:

https://sphinx-rtd-theme.readthedocs.io/en/3.0.1/installing.html


joshuacwnewton (approved):
LGTM! (With the Python 3.7 error accounted for by https://github.com/spine-generic/spine-generic/pull/280.)

PR #229, opened 2020-09-29
sg_create_mosaic: Fix issues with modules compatibility; various improvements

alexfoias:
Fixes https://github.com/spine-generic/spine-generic/issues/225

To try (on Compute Canada):
```
module load python/3.7
# activate venv
# git checkout -b af/fix_create_mosaic origin/af/fix_create_mosaic  # do it once
git checkout af/fix_create_mosaic
pip install -e .
sg_create_mosaic -i *T1w_RPI_r_flatten.nii.gz -ifolder ~/project/results_multi_20200907/data_processed/ -s _seg.nii.gz -p sag -col 20 -row 12 -o fig_mosaic_T1w.png
```

PR #280, opened 2024-10-09
Update supported Python versions to `">=3.9"`

### LINKED ISSUE #242: Support other python versions
For the moment spine-generic only supports python 3.7.
We would like to extend it to 3.8 and 3.9

mguaypaq:
Python 3.7 officially reached end-of-life in June 2023, and 3.8 reached end-of-life two days ago. Meanwhile, 3.12 and 3.13 are the current most-supported versions.

See: https://devguide.python.org/versions/

Fixes https://github.com/spine-generic/spine-generic/issues/242.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ statsmodels
 pyyaml
 pybids
 scikit-image
-transforms3d~=0.3.1
+transforms3d
 pandas-schema
 plotly~=4.12.0
 opencv-python
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,6 @@ statsmodels
 pyyaml
 pybids
 scikit-image
-transforms3d
 pandas-schema
 plotly~=4.12.0
 opencv-python
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ with open(req_path, "r") as f:
 setup(
     name="spinegeneric",
     version=spinegeneric.__version__,
-    python_requires=">=3.7,<3.12",
+    python_requires=">=3.9,<=3.13",
     description="Collection of cli to process data for the Spine Generic project.",
     url="https://spine-generic.rtfd.io",
     author="NeuroPoly Lab, Polytechnique Montreal",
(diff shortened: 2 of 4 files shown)


mguaypaq:
I also removed the Python 3.7 and 3.8 tests from the required checks in the `master` branch protection rules in the repository settings. I didn't add 3.12 and 3.13, because it looks like they're currently failing (do we want to be testing these newer versions at all?)

joshuacwnewton:
> I didn't add 3.12 and 3.13, because it looks like they're currently failing (do we want to be testing these newer versions at all?)

I think it depends on how easy the tests are to fix? If the failing tests have simple solutions, then sure! But if there's some major blocker, then... it's not the highest priority issue, aha.

I see that we're pinning `transforms3d~=0.3.1`, which I think prevents us from getting the newest version ([0.4.2](https://pypi.org/project/transforms3d/0.4.2/)). I might try pushing a commit just to see what happens when we unpin this? But, I won't hold up the PR with more failures either.

requirements.txt
```diff
@@ -9,7 +9,7 @@ statsmodels
 pyyaml
 pybids
 scikit-image
-transforms3d~=0.3.1
+transforms3d
```
joshuacwnewton, line 12: Blaming `requirements.txt`, it looks like `transforms3d` has been pinned from the start (https://github.com/spine-generic/spine-generic/pull/229). And, it looks like this pin was originally supposed to target the _minimum_ version (https://github.com/spine-generic/spine-generic/pull/229/commits/798763ca92be0a3c923fa6c6241abd5d73d75bbe).

I think this means that it's probably fine to remove this pin? `transforms3d` seems to update [relatively infrequently](https://github.com/matthew-brett/transforms3d/issues/46), too, so I imagine the auto-updates would mainly be when something breaks upstream and the author is releasing a fix.

joshuacwnewton (approved):

requirements.txt
mguaypaq: Thanks for digging! I think upgrading `transforms3d` looks safe enough. But it looks like it's never directly imported in `spine-generic`, only used (possibly?) indirectly through dependencies. In which case, without a version pin, maybe we don't have to mention `transforms3d` at all?

mguaypaq, line 12:
```suggestion
```

setup.py
```diff
@@ -18,7 +18,7 @@
 setup(
     name="spinegeneric",
     version=spinegeneric.__version__,
-    python_requires=">=3.7,<3.12",
+    python_requires=">=3.9,<=3.13",
```
mguaypaq, line 21: I recently happened to see [a giant post against upper version bounds](https://iscinumpy.dev/post/bound-version-constraints/) in `python_requires`, so:
```suggestion
    python_requires=">=3.9",
```

mguaypaq:
Thanks for the extra fixes @joshuacwnewton!

PR #285, opened 2025-05-19
Update Python packaging format

mguaypaq:
This is a slight overhaul of the Python package structure, to reflect more modern standards.

I'll make more substantial changes to the code in follow-up pull requests.


mguaypaq:
I switched from declaring the dependencies in `requirements.txt` to `pyproject.toml` because we weren't pinning most of the versions, but if this causes trouble in the future, I'd be very open to reinstating a version lock file listing a set of exact versions that are known to work.

pyproject.toml
joshuacwnewton, line 1: I was wondering how the new `pyproject.toml` was generated -- by hand, or automatically from a conversion tool?

So, I was curious how this file would compare to the output of an tool, so I ran `hatch new --init`, and the outputs were [effectively the same](https://www.diffchecker.com/eMZummnl/) (excluding the hatch-specific stuff).

pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+dev = [
+  "flake8",
+]
+docs = [
+  "sphinx-rtd-theme~=3.0.1",
+]
+
+[project.urls]
+Homepage = "https://spine-generic.readthedocs.io/"
+Repository = "https://github.com/spine-generic/spine-generic"
```
joshuacwnewton, line 54: Thank you for getting rid of the `rtfd` link that was present in the old `setup.py`. That always struck me as being in bad taste. :sweat_smile:

src/spinegeneric/cli/generate_figure.py
```diff
@@ -1343,7 +1340,3 @@ def main(argv=sys.argv[1:]):
     if args.output_html:
         # Generate interactive html T1w vs. T2w figure
         generate_figure_t1_t2_plotly(df, csa_t1, csa_t2)
-
-
-if __name__ == "__main__":
-    main()
```
joshuacwnewton, line 1349: Regarding 

> * Remove `if __name__ == "__main__":` footers
>  (The CLI scripts are intended to be called via the pip-installed entry
>  points `sg_something`, rather than `python -m spinegeneric.something`.)

I assume the only reason SCT needs the `__name__` footers is because our entry points call a launcher script which then calls the `sct_` scripts using a subprocess? Whereas here, the `sg_` entry points tap directly into `main()`, making it unnecessary...

src/spinegeneric/cli/qc_bids_deface.py
```diff
@@ -1,97 +1,94 @@
...
-    path_qc_report = output_path + "/qc_report.html"
-    html_file = open(output_path + "/qc_report.html", "w")
-    html_file.write(final_qc)
-    html_file.close()
-
-command = """open -a "Google Chrome" """ + path_qc_report
-os.system(command)
+
+
+def main():
```
joshuacwnewton, line 7: Wait, `qc_bids_deface` just straight up didn't have a `main()` function despite its entry point being:

```
sg_qc_bids_deface = "spinegeneric.cli.qc_bids_deface:main"
```

Does that mean this script has just been... broken until now?

pyproject.toml
joshuacwnewton: I guess one difference is in the `classifiers` -- I see it specifies 3.12 and 3.13 as well as MIT, while the PR's only specifies up to 3.11.

I guess we are currently including 3.12 and 3.13 in sg's test suite... maybe the classifiers are worth including?

setup.py
```diff
@@ -1,58 +0,0 @@
...
-
-req_path = path.join(here, "requirements.txt")
-with open(req_path, "r") as f:
-    install_reqs = f.read().strip()
-    install_reqs = install_reqs.split("\n")
-
-setup(
-    name="spinegeneric",
-    version=spinegeneric.__version__,
-    python_requires=">=3.9",
```
joshuacwnewton, line 21: Apologies if I'm being dense here, but if the old `setup.py` supported Python 3.9, in what way is this PR upgrading the Python version?

joshuacwnewton (approved):
These changes seem very reasonable and straightforward! I hardly have any changes to request at all, mostly just comments/questions.

pyproject.toml
mguaypaq: I generated it by hand, mostly following these documentations pages:
- [from packaging.python.org](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/)
- [from `setuptools`](https://setuptools.pypa.io/en/latest/userguide/index.html)
- [from the `setuptools-scm` plugin](https://setuptools-scm.readthedocs.io/en/stable/usage/).

For the license classifier, I think `setuptools` has a hard deprecation, so I removed it. For the Python version classifiers, I figured I'd wait to include 3.12 and 3.13 since the jump from 3.11 to 3.12 has a bunch of deprecations and API changes that I haven't considered yet.

Honestly, I considered only including the generic `Programming Language :: Python :: 3` classifier, because:
- The only impact of classifiers is for browsing and searching on PyPI. They're not taken into account by `pip` at all.
- We don't expect anyone to search for this package on PyPI anyway? And certainly not by filtering on Python versions?
- In fact, we're not putting this package on PyPI at all (intentionally so, because I don't think NeuroPoly as an organisation is willing to put in the maintenance work for PyPI, especially since `pip` can already install from github).

src/spinegeneric/cli/generate_figure.py
mguaypaq: That's right!

src/spinegeneric/cli/qc_bids_deface.py
mguaypaq: I'm pretty sure it was, yes. (Or else, it accidentally worked because all of its code ran on `import` and it prompted for its arguments interactively, before the launcher script could error out on the lack of a `main` function to call.)

Honestly, I think some of the code in this repo was last run a few years ago, when the paper was being prepared.

setup.py
mguaypaq: You're absolutely right, it's not! I think I got confused because the Github Actions for this repo, and for data-single-subject, and for data-multi-subject, all pin different releases of the repo. And at least one of them used to fail because Python 3.7 was no longer supported. I'll update the PR title.

pyproject.toml
mguaypaq: Anyway, I added 3.12 and 3.13! So that's done.

PR #265, opened 2023-04-11
Clean up `sg_params_checker` so it can be run on `data-multi-subject` without crashing

joshuacwnewton:
## Description

This PR has 3 parts:

- The first 16 commits are refactoring commits to make the script clearer and more concise. They clean up the code style of the script, while also adding documentation to make certain aspects (`BIDSLayout`, `BIDSLayoutIndexer`) easier to understand.
- The second to last commit, dc73e639e714ce4c64286b1b013d92386a797954, fixes #264.
- The last commit, 4adaeb7f09dfe5e66a2cb940a25ceccf1a4a88a4, adds some TODOs for discussion in this PR. 

Running the script on `data-multi-subject` results in 142 warnings:

<details>

```
WARNING: sub-amu01_T2w.nii.gz: Incorrect RepetitionTime: TR=2 instead of 1.5 +/- 0.1.
WARNING: sub-amu01_T2w.nii.gz: Incorrect FlipAngle: FA=180 instead of 120.
WARNING: sub-amu02_T2w.nii.gz: Incorrect RepetitionTime: TR=2 instead of 1.5 +/- 0.1.
WARNING: sub-amu02_T2w.nii.gz: Incorrect FlipAngle: FA=180 instead of 120.
WARNING: sub-amu03_T2w.nii.gz: Incorrect RepetitionTime: TR=2 instead of 1.5 +/- 0.1.
WARNING: sub-amu03_T2w.nii.gz: Incorrect FlipAngle: FA=135 instead of 120.
WARNING: sub-amu04_T2w.nii.gz: Incorrect RepetitionTime: TR=2 instead of 1.5 +/- 0.1.
WARNING: sub-amu04_T2w.nii.gz: Incorrect FlipAngle: FA=180 instead of 120.
WARNING: sub-amu05_T2w.nii.gz: Incorrect RepetitionTime: TR=2 instead of 1.5 +/- 0.1.
WARNING: sub-amu05_T2w.nii.gz: Incorrect FlipAngle: FA=180 instead of 120.
WARNING: sub-beijingVerio01_T2w.nii.gz: Incorrect RepetitionTime: TR=2 instead of 1.5 +/- 0.1.
WARNING: sub-beijingVerio01_T2w.nii.gz: Incorrect FlipAngle: FA=180 instead of 120.
WARNING: sub-beijingVerio02_T2w.nii.gz: Incorrect RepetitionTime: TR=2 instead of 1.5 +/- 0.1.
[...]

PR #286, opened 2025-05-21
Revamp the `sg_params_checker` code

mguaypaq:
~I still want to look at the TODO comments around `BIDSLayout`~ (done), but here are a bunch of small changes already.


src/spinegeneric/utils.py
```diff
@@ -40,7 +40,6 @@ class SmartFormatter(argparse.HelpFormatter):
     """
 
     def __init__(self, *args, **kw):
-        self._add_defaults = None
```
joshuacwnewton, line 43: Embarrassing! Looking at the attribution link included in the docstring for this class, we based our formatter off of `skaff.cli` that seems to derive from `ruamel.std.argparse`, which indeed contains this line + [an actual use of `_add_defaults`](https://sourceforge.net/p/ruamel-std-argparse/code/ci/default/tree/__init__.py#l114):

```python
    def _get_help_string(self, action):
        if self._add_defaults is None:
            return argparse.HelpFormatter._get_help_string(self, action)
        help = action.help
        if '%(default)' not in action.help:
            if action.default is not argparse.SUPPRESS:
                defaulting_nargs = [argparse.OPTIONAL, argparse.ZERO_OR_MORE]
                if action.option_strings or action.nargs in defaulting_nargs:
                    help += ' (default: %(default)s)'
        return help
```

We definitely don't have this (anymore?)!

src/spinegeneric/cli/params_checker.py
```diff
@@ -13,22 +13,20 @@
...
-import spinegeneric.utils
 
 
 def get_parser():
     parser = argparse.ArgumentParser(
-        description="Acquisition parameters checker feature. This feature allows the users "
-                    "to compare the acquisition parameters that can be found in the json "
-                    "sidecar to the recommended acquisition parameters.",
-        formatter_class=spinegeneric.utils.SmartFormatter,
-        prog=os.path.basename(__file__).strip(".py"),
```
joshuacwnewton: I'm curious why we dropped these lines in commit b50093673845d63548c966afe7f49e1f13c42069. They seem to still be used in [all of the other CLI entry points](https://github.com/search?q=repo%3Aspine-generic%2Fspine-generic%20smartformatter&type=code), so I'm curious if we plan to remove them from `spine-generic` as a whole, or if they were just not useful for `params_checker` in particular.

src/spinegeneric/cli/params_checker.py
```diff
@@ -8,7 +8,8 @@
 import importlib.resources
 import json
 import logging
-import os
+import logging.config
```
joshuacwnewton, line 11: I notice that commit 6f0d60afe39dbf91ead65c96f4f74d7006fbd450 added an extra import of `logging.config`. Is it fair to presume that that change was meant for https://github.com/spine-generic/spine-generic/pull/286/commits/b563ecc69997dcbfc521bf679aec7a9bb18e0e38, due to the use of `logging.config.dictConfig` (which is one level deeper)?

If so, then no need to rewrite the commits or anything, just wanted to double-check. :)

mguaypaq:
Some investigation today about these lines:
https://github.com/spine-generic/spine-generic/blob/a736b8424a9acbab64df5b86c0237adcbf5000ce/src/spinegeneric/cli/params_checker.py#L70-L76

TLDR: We should just get rid of our custom configuration. I will add a commit for this.

1. Commenting out this line had no effect on the output when run in `data-single-subject` or `data-multi-subject`. But since `data-multi-subject` takes longer to run, I limited my further tests to just `data-single-subject`.
2. But, looking at [the documentation for `config_filename`](https://github.com/bids-standard/pybids/blob/50656a827917489c729ebf3072e4b88efbcfd530/src/bids/layout/index.py#L104-L106
) made me think we're using the wrong parameter, and maybe we're just _not using our custom config file anyway_.

   In fact, that parameter is meant to be a generic name (default `layout_config.json`) for files that can be included at various levels of the dataset to _add extra rules_ for a subtree of the dataset (kind of like `.gitignore`). This is confirmed by looking at [the code using this parameter](https://github.com/bids-standard/pybids/blob/50656a827917489c729ebf3072e4b88efbcfd530/src/bids/layout/index.py#L199-L204).

   Instead, we're passing an absolute path, so either:
   - this is getting completely ignored, or
   - we're actually loading _extra_ allowed patterns _for every directory being scanned_.

   Either way, this can't be what we want.
3. The correct way is to drop the custom `BIDSLayoutIndexer`, and simply pass `path_sg_layout_config` to the `config` parameter of `BIDSLayout`.
4. This has an unexpected effect: more warnings appear in the output!
   - One would hope/fear that this is because we've been silently neglecting to check some files.
   - But in fact, the new warnings are _incorrect_: they're based on the wrong metadata for various image files.
     - Here's some output showing that `BIDSLayout` with the default config finds the right metadata files:

       ```py
       >>> print(some_file.relpath)
       sub-chiba750/anat/sub-chiba750_flip-2_mt-off_MTS.nii.gz
       >>> for json in some_file.get_associations(include_parents=True): print(json.relpath)
       ...
       sub-chiba750/anat/sub-chiba750_flip-2_mt-off_MTS.json
       ```
     - And here's some output showing that the custom config gives too many metadata files:

       ```py
       >>> print(some_file.relpath)
       sub-chiba750/anat/sub-chiba750_flip-2_mt-off_MTS.nii.gz
       >>> for json in some_file.get_associations(include_parents=True): print(json.relpath)
       ...
       sub-chiba750/anat/sub-chiba750_flip-1_mt-off_MTS.json
       sub-chiba750/anat/sub-chiba750_flip-1_mt-on_MTS.json
       sub-chiba750/anat/sub-chiba750_flip-2_mt-off_MTS.json
       sub-chiba750/anat/sub-chiba750_flip-1_mt-on_MTS.nii.gz
       sub-chiba750/anat/sub-chiba750_flip-2_mt-off_MTS.nii.gz
       ```
5. This is the result of [the weird "inheritance principle" in BIDS](https://bids-specification.readthedocs.io/en/stable/common-principles.html#the-inheritance-principle), together with our custom config's _lack_ of definition for the `flip` and `mt` entities.

   Since our custom config doesn't define these entities, `BIDSLayout` ignores those parts of the filenames, so they all look like just `{"sub": "chiba750", "suffix": "MTS"}`. Since they have the same entities, they're considered to all refer to the same image (but they don't).

In other words, because of many subtle interactions, one of our bugs was neutralizing another one of our bugs.

mguaypaq:
Also, as to the original source of our custom configuration, see https://github.com/spine-generic/spine-generic/pull/265#discussion_r1184118938. The conclusion is that our initial additions to the config have since been integrated upstream, so they're obsolete.

mguaypaq:
Re: this other point:

https://github.com/spine-generic/spine-generic/blob/a736b8424a9acbab64df5b86c0237adcbf5000ce/src/spinegeneric/cli/params_checker.py#L77-L84

I'm not sure that upstream documentation is entirely consistent and up-to-date, but empirically there's no change in the output, and there are no relevant files under `derivatives/` for either `data-single-subject` or `data-multi-subject`.

I would say: let's keep `validate=False`, because
1. I'd rather not ignore files silently.
2. I'd rather not do extra validation work that we don't care about.

src/spinegeneric/cli/params_checker.py
```diff
@@ -106,24 +106,40 @@ def main():
...
             continue
 
-        # Parse the file's contrast from its suffix (sans `.nii.gz`)
-        Contrast = (item.filename.split("_")[-1]).split(".")[0]
-        # In the case of MTS files, the spine-generic protocol doesn't just specify 'MTS'. Instead, 'MTon_MTS',
-        # 'MToff_MTS', 'T1w_MTS' etc. are used. So, we need to parse the type of MTS from the filename,
-        # then convert it to the specific names expected by the 'manufacturer params' dictionary.
+        # Parse the filename's BIDS entities and suffix
+        parts = item.filename.removesuffix(".nii.gz").split("_")
+        suffix = parts.pop()
```
joshuacwnewton, line 110: I notice that the old code (inadvertently) would have handled file extensions other than `.nii.gz` (by discarding everything after to the first `.` via `.split(".")[0]`). For example, with the new code, I think a `.nii` suffix would get parsed as a contrast instead? 

Maybe it would be worth doing something like this instead using [`suffixes`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffixes)?

```python
        # Parse the filename's BIDS entities and suffix
        extension = "".join(item.suffixes)
        if extension not in [".nii", ".nii.gz"]:
            logging.warning(f"{item.filename}: Non-NIfTI file extension encountered: {extension}")
        parts = item.filename.removesuffix(extension).split("_")
        suffix = parts.pop()
```

That way, the old functionality is preserved (removing the full extension) while also giving a warning if an unexpected extension is encountered.

joshuacwnewton: Though, maybe we can safely assume everything will be `.nii.gz` due to some sort of previous file filtering? I wouldn't be surprised if we had already selected only `.nii.gz` files earlier in the script.

src/spinegeneric/cli/params_checker.py
```diff
@@ -141,31 +141,40 @@ def main():
...
-                logging.warning(f"{item.filename}: Incorrect FlipAngle: "
-                                f"FA={FlipAngle} instead of {ExpectedFA}.")
+        # Fetch the available parameters for the given manufacturer + model
+        expected = sg_acq_protocol[Manufacturer][ManufacturersModelName][Contrast]
+
+        # Validate values against spine-generic's acquisition protocol
+        for key, symbol, tolerance in [
+            ("RepetitionTime", "TR", 0.1),
+            ("EchoTime", "TE", 0.1),
+            ("FlipAngle", "FA", None),
```
joshuacwnewton, line 151: Tiny nitpick, but would setting `tolerance` to `0.0` do the same job as the exact comparison code?

(I guess it causes a float comparison rather than a (default filetype) comparison, which may be unwanted...)

src/spinegeneric/cli/params_checker.py
```diff
@@ -58,6 +58,9 @@ def main():
         "root": {"level": "DEBUG", "handlers": ["console", "file"]},
     })
 
+    # Keep track of whether any warnings are found
+    warnings_found = False
```
joshuacwnewton, line 62: Instead of updating `warnings_found` everywhere we throw a warning, could we instead add a third handler (`StreamHandler`) that sends logs to a `StringIO` stream object, then use `stream.getvalue()` and check if any warnings were logged when deciding whether to emit 1 or 0?

src/spinegeneric/cli/params_checker.py
```diff
@@ -64,25 +64,7 @@ def main():
...
     #   "A BIDSLayout instance is a lightweight container for all files in the BIDS project directory."
-    with importlib.resources.path(spinegeneric.config, "bids_specs.json") as path_sg_layout_config:
-        layout = BIDSLayout(
-            str(args.path_in),
-            # BIDSLayoutIndexer is a class that indexes files based on pattern-matching defined in the config.
-            # By default, BIDS has its own config. But, SG specifies its own custom config instead. (Why?)
-            # TODO: The default config fetches 1573 files from data-multi-subject, but the modified config
-            #       *also* fetches 1573 files. Do they always perform identically? In what cases is the custom
-            #       config even needed? It would be nice to add comments to `bids_specs.json` to highlight the
-            #       areas where the custom config deviates from the built-in, default config.
```
joshuacwnewton, line 66: Thank you so much for investigating my (years-old, now!) comments. I'm glad to see that my suspicions were on the right track, and happy to see our old config get tossed.

joshuacwnewton (changes requested):
The changes mostly LGTM! I just had some nitpicky tweaks/comments/questions, but the bones of the PR are very good. :hearts:

src/spinegeneric/cli/params_checker.py
mguaypaq: I would like to remove them generally, but for this PR I'm focusing on just `sg_params_checker`. And in the next PR I want to focus on `sg_check_data_consistency`. These are the only two scripts that we run in CI for `data-single-subject` and `data-multi-subject`. After that, I'd be happy to do some minor cleanups like this in the other scripts (and otherwise let sleeping dogs lie).

The reasons:
- I think the `SmartFormatter` is kind of a hack (personal opinion), and it's mostly obsolete (the wrap-to-terminal width feature is now present upstream), and it does line-wrapping poorly (it cuts the line mid-word, unlike upstream).
- The `.strip(".py")` thing is actually a bug in all the scripts. The intent is to remove the `.py` suffix in the program name, but:
  - The default `prog` would be the entry point `sg_params_checker` (with no `.py` suffix to begin with), whereas `os.path.basename(__file__)` is the module name `params_checker.py`, which is missing the `sg_` prefix.
  - The effect of `.strip(".py")` is to remove all instances of the three characters `.`, `p`, `y` from each end, resulting in the (multiply) incorrect result `arams_checker`. If anything, `.removesuffix(".py")` should have been used.

src/spinegeneric/cli/params_checker.py
mguaypaq: Exactly!

src/spinegeneric/cli/params_checker.py
mguaypaq: I considered doing something like this, but since the earlier code explicitly filters the `query` on `extension="nii.gz"`, I decided to keep this simpler code.

src/spinegeneric/cli/params_checker.py
mguaypaq: Yes, this bothered me a bit. I would probably do it with a custom handler class that just holds a boolean and turns it to `True` when receiving messages. But:
- The problem is that we're currently using the root handler, and the `bids` module's various messages also end up there. We'd have to set up our own non-root handler, and hope that no-one else uses it (extremely unlikely, but still). I didn't really feel like overhauling that part of the code.
- Another extremely unlikely worry is that I think `logging` allows a lot of configuration hooks (the idea is that the user, or some other part of a system that import us, would want to override some of the configuration), which might interfere with this non-logging function of returning the right exit code for the script. So, in some philosophical sense, I think the logging infrastructure is the "wrong" source for a separate functionality like the script's exit code.

How about I factor this out into a function which calls `logging.warn` and sets the boolean? That feels like the right way around to me.

src/spinegeneric/cli/params_checker.py
mguaypaq: That would result in two unwanted consequences:
- Type conversion to `float` (whereas the current code would allow checking a string parameter, for instance).
- Adding an extra `+/- 0.0` at the end of the warning message.

src/spinegeneric/cli/params_checker.py
joshuacwnewton: Ah!! Okay, I think I know why this faulty code exists in the first place -- SCT has its `launcher.py` entry point, which then calls the Python files directly, meaning that we had an extraneous `.py` in all of our `prog` entries.

Then, presumably we just copied and pasted without adjusting it for the (now direct-to-script) new entry points.

src/spinegeneric/cli/params_checker.py
joshuacwnewton: Ahhhh, that makes perfect sense. Sure!

joshuacwnewton (approved):
Based on further discussion, and assuming only a minor change for the `warnings_found = ` nitpick, this LGTM!

src/spinegeneric/cli/params_checker.py
mguaypaq: Done! How about this?

joshuacwnewton: Perfect! :hearts:

mguaypaq:
Thanks for the review!

PR #287, opened 2025-05-23
Revamp the `sg_check_data_consistency` code

mguaypaq:
Here are a few changes ~so far. More to come~.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,6 @@ dependencies = [
   "nibabel",
   "opencv-python",
   "pandas",
-  "pandas-schema",
   "plotly~=4.12.0",
   "pybids",
   "pyyaml",
--- a/src/spinegeneric/cli/check_data_consistency.py
+++ b/src/spinegeneric/cli/check_data_consistency.py
@@ -125,13 +125,15 @@ def main():
         if not json_path.exists():
             warn(f"missing JSON sidecar for {img_path}")
 
-    # Check the contents of participants.tsv
+    # Check the column names of `participants.tsv` to ensure they match the validator
     tsv_cols = set(fieldnames)
     expected_cols = set(validators.keys())
     for col in sorted(expected_cols - tsv_cols):
         warn(f"participants.tsv: missing column '{col}'")
     for col in sorted(tsv_cols - expected_cols):
         warn(f"participants.tsv: extra column '{col}'")
+
+    # Check the row values of `participants.tsv` using the validator
     for r, row in enumerate(rows, start=1):
         if None in row.values():
             warn(f"participants.tsv: row {r} is too short")
--- a/src/spinegeneric/cli/check_data_consistency.py
+++ b/src/spinegeneric/cli/check_data_consistency.py
@@ -6,6 +6,7 @@ For usage, type: sg_check_data_consistency -h
 
 import argparse
 import os
+from pathlib import Path
 from pprint import pprint
 
 import pandas as pd
@@ -19,23 +20,20 @@ from pandas_schema.validation import (
     TrailingWhitespaceValidation,
 )
 
-import spinegeneric as sg
-import spinegeneric.utils
-
 
 def get_parser():
     parser = argparse.ArgumentParser(
-        description="Data consistency checker feature. This feature allows the users"
-        "to check the subjects listed in participants.tsv and the actual sub-* data."
-        "In addition, it checks the presence of jsonSidecar.",
-        formatter_class=sg.utils.SmartFormatter,
-        prog=os.path.basename(__file__).strip(".py"),
+        description="""
+            Script to check the contents of 'participants.tsv', compare it
+            against the set of 'sub-*' folders, and check the presence of JSON
+            sidecars.
+        """,
     )
     parser.add_argument(
         "-path-in",
         required=True,
-        type=str,
-        help="Path to input BIDS dataset, which contains all the 'sub-' folders.",
+        type=Path,
+        help="Path to input BIDS dataset, which contains all the 'sub-*' folders.",
     )
     return parser
 
@@ -45,7 +43,7 @@ def main():
     parser = get_parser()
     args = parser.parse_args()
 
-    data_path = args.path_in
+    data_path = str(args.path_in)
 
     path_tsv = os.path.join(data_path, "participants.tsv")
     tsv_file = pd.read_csv(path_tsv, sep="\t")
--- a/src/spinegeneric/cli/check_data_consistency.py
+++ b/src/spinegeneric/cli/check_data_consistency.py
@@ -99,6 +99,14 @@ def main():
     parser = get_parser()
     args = parser.parse_args()
 
+    # Keep track of errors for the exit status
+    warnings_found = False
+
+    def warn(message: str):
+        nonlocal warnings_found
+        warnings_found = True
+        print(f"Warning: {message}")
+
     # Read participants.tsv
     fieldnames, rows = read_tsv(args.path_in / "participants.tsv")
 
@@ -107,28 +115,28 @@ def main():
         tsv_subj = set(r["participant_id"] for r in rows)
         dir_subj = set(p.name for p in args.path_in.glob("sub-*") if p.is_dir())
         for subj in sorted(dir_subj - tsv_subj):
-            print(f"Warning: participants.tsv: missing row for data folder '{subj}'")
+            warn(f"participants.tsv: missing row for data folder '{subj}'")
         for subj in sorted(tsv_subj - dir_subj):
-            print(f"Warning: participants.tsv: missing data folder for subject '{subj}'")
+            warn(f"participants.tsv: missing data folder for subject '{subj}'")
 
     # Check the presence of JSON sidecars
     for img_path in args.path_in.glob("sub-*/**/*.nii.gz"):
         json_path = img_path.with_suffix("").with_suffix(".json")
         if not json_path.exists():
-            print(f"Warning: missing JSON sidecar for {img_path}")
+            warn(f"missing JSON sidecar for {img_path}")
 
     # Check the contents of participants.tsv
     tsv_cols = set(fieldnames)
     expected_cols = set(validators.keys())
     for col in sorted(expected_cols - tsv_cols):
-        print(f"Warning: participants.tsv: missing column '{col}'")
+        warn(f"participants.tsv: missing column '{col}'")
     for col in sorted(tsv_cols - expected_cols):
-        print(f"Warning: participants.tsv: extra column '{col}'")
+        warn(f"participants.tsv: extra column '{col}'")
     for r, row in enumerate(rows, start=1):
         if None in row.values():
-            print(f"Warning: participants.tsv: row {r} is too short")
+            warn(f"participants.tsv: row {r} is too short")
         if None in row.keys():
-            print(f"Warning: participants.tsv: row {r} is too long")
+            warn(f"participants.tsv: row {r} is too long")
         for col, validate in validators.items():
             if col not in fieldnames:
                 continue
@@ -138,4 +146,7 @@ def main():
             try:
                 validate(value)
             except ValueError as e:
-                print(f"Warning: participants.tsv: row {r}: {col} is '{value}', but {e}")
+                warn(f"participants.tsv: row {r}: {col} is '{value}', but {e}")
+
+    # exit code
+    return 1 if warnings_found else 0
--- a/src/spinegeneric/cli/check_data_consistency.py
+++ b/src/spinegeneric/cli/check_data_consistency.py
@@ -5,7 +5,6 @@ For usage, type: sg_check_data_consistency -h
 """
 
 import argparse
-import os
 from pathlib import Path
 from pprint import pprint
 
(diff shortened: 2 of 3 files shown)


mguaypaq:
I'm looking into our use of `pandas_schema`, and this is what I've found so far:
- For `data-single-subject`, we have too few columns compared to the schema declared in `sg_check_data_consistency`, so the only output is:

  ```
  Invalid number of columns. The schema specifies 13, but the data frame has 7
  ```
- For `data-multi-subject`, we have some extra columns compared to the schema, so the only output is:

  ```
  Invalid number of columns. The schema specifies 13, but the data frame has 15
  ```
- The upstream repository is now archived: https://github.com/multimeric/PandasSchema.

Given that our validation is pretty simple, I'm tempted to just remove this dependency and iterate over the rows of the tsv file.

mguaypaq:
With this latest commit, I think I get reasonable output on `data-single-subject`:

```
(env9) mguaypaq@torme:~/spine-generic/data-single-subject$ sg_check_data_consistency -path-in .
Warning: participants.tsv: missing column 'height'
Warning: participants.tsv: missing column 'institution'
Warning: participants.tsv: missing column 'receive_coil_name'
Warning: participants.tsv: missing column 'researcher'
Warning: participants.tsv: missing column 'software_versions'
Warning: participants.tsv: missing column 'weight'
Warning: participants.tsv: row 1: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 2: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 3: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 4: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 5: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 6: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 7: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 8: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 9: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 10: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 11: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 12: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 13: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 14: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 15: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 16: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 17: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 18: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
Warning: participants.tsv: row 19: sex is 'm', but should be one of 'M', 'F', 'O', or 'n/a'
(env9) mguaypaq@torme:~/spine-generic/data-single-subject$ echo $?
1
```

And on `data-multi-subject`:

```
(env9) mguaypaq@torme:~/spine-generic/data-multi-subject$ sg_check_data_consistency -path-in .
Warning: participants.tsv: extra column 'notes'
Warning: participants.tsv: extra column 'pathology'
Warning: participants.tsv: row 166: weight is '83.91', but should be an integer
(env9) mguaypaq@torme:~/spine-generic/data-multi-subject$ echo $?
1
```

mguaypaq:
I considered switching from `print()`-style output to `logging.warning()`-style output, like in `sg_params_checker`, but I didn't really see any benefit.

src/spinegeneric/cli/check_data_consistency.py
```diff
@@ -68,12 +60,10 @@ def main():
...
         pprint(missing_subjects_folder)
 
-    for dirName, subdirList, fileList in os.walk(data_path):
-        for file in fileList:
-            if file.endswith(".nii.gz"):
-                jsonSidecarPath = os.path.join(dirName, file.split(".")[0] + ".json")
-                if not os.path.exists(jsonSidecarPath):
-                    print("Missing jsonSidecar: " + jsonSidecarPath)
+    for img_path in args.path_in.glob("sub-*/**/*.nii.gz"):
+        json_path = img_path.with_suffix("").with_suffix(".json")
```
joshuacwnewton, line 124: Interesting! Chaining calls to `.with_suffix` to replace the suffix... :writing_hand: :thinking:

src/spinegeneric/cli/check_data_consistency.py
```diff
@@ -37,29 +37,33 @@ def get_parser():
     return parser
 
 
+Fieldname = str
+Row = dict[Fieldname, str]
```
joshuacwnewton, line 32: Huh! This is a typing pattern I hadn't seen before. I like this a lot? A neat way to add more contextual information to objects represented by stdlib types, without needing to make a completely new class for typing purposes.

src/spinegeneric/cli/check_data_consistency.py
```diff
@@ -111,6 +115,7 @@ def main():
         ]
     )
 
+    tsv_file = pd.read_csv(str(args.path_in / "participants.tsv"), sep="\t")
```
joshuacwnewton: Note to self: We still use `pandas` for `schema.validate`, double-check if this gets removed later on.

src/spinegeneric/cli/check_data_consistency.py
```diff
@@ -47,6 +38,62 @@ def read_tsv(path: Path) -> tuple[list[Fieldname], list[Row]]:
...
+def validate_sex(value: str):
+    validate_whitespace(value)
+    if value not in ["M", "F", "O"]:
+        raise ValueError("should be one of 'M', 'F', 'O', or 'n/a'")
+
+
+def validate_age(value: str):
+    age = validate_integer(value)
+    if not 18 <= age <= 60:
+        raise ValueError("should be between 18 and 60")
```
joshuacwnewton, line 65: Is there any merit to generalizing these to be functions that return function objects, such that we can specify configuration arguments (and thus make the `validators` dictionary more self-documented)?

```python
    "sex": validate_list(["M", "F", "O"]),
    "age": validate_range([18, 60]),
```

Kind of like how the old schema functionality was supposed to work?

src/spinegeneric/cli/check_data_consistency.py
```diff
@@ -56,70 +103,39 @@ def main():
...
+            print(f"Warning: missing JSON sidecar for {img_path}")
+
+    # Check the contents of participants.tsv
+    tsv_cols = set(fieldnames)
+    expected_cols = set(validators.keys())
+    for col in sorted(expected_cols - tsv_cols):
+        print(f"Warning: participants.tsv: missing column '{col}'")
+    for col in sorted(tsv_cols - expected_cols):
+        print(f"Warning: participants.tsv: extra column '{col}'")
+    for r, row in enumerate(rows, start=1):
```
joshuacwnewton: Tiny organizational nit to avoid a "wall of text" feel:

```suggestion
    # Check the column names of `participants.tsv` to ensure they match the validator
    tsv_cols = set(fieldnames)
    expected_cols = set(validators.keys())
    for col in sorted(expected_cols - tsv_cols):
        print(f"Warning: participants.tsv: missing column '{col}'")
    for col in sorted(tsv_cols - expected_cols):
        print(f"Warning: participants.tsv: extra column '{col}'")
        
    # Check the row values of `participants.tsv` using the validator
    for r, row in enumerate(rows, start=1):
```

src/spinegeneric/cli/check_data_consistency.py
```diff
@@ -56,70 +103,39 @@ def main():
...
-            print(f"Missing jsonSidecar: {json_path}")
-
-    # Checking participants.tsv contents
-    schema = Schema(
-        [
-            Column(
-                "participant_id",
-                [LeadingWhitespaceValidation(), TrailingWhitespaceValidation()],
-            ),
-            Column("sex", [InListValidation(["M", "F"])]),
```
joshuacwnewton, line 85: I'm curious about the change from `["M", "F"]` to `["M", "F", "O"]`. Is Other part of the data spec? :D

src/spinegeneric/cli/check_data_consistency.py
```diff
@@ -47,6 +38,62 @@ def read_tsv(path: Path) -> tuple[list[Fieldname], list[Row]]:
...
+    try:
+        return int(value)
+    except ValueError:
+        raise ValueError("should be an integer") from None
+
+
+def validate_sex(value: str):
+    validate_whitespace(value)
+    if value not in ["M", "F", "O"]:
+        raise ValueError("should be one of 'M', 'F', 'O', or 'n/a'")
```
joshuacwnewton, line 59: I notice the use of `n/a` in the error message, but missing from the check itself. Is this intentional? (i.e. will `n/a` get parsed as `None` and/or skipped?)

src/spinegeneric/cli/check_data_consistency.py
```diff
@@ -99,6 +99,14 @@ def main():
     parser = get_parser()
     args = parser.parse_args()
 
+    # Keep track of errors for the exit status
+    warnings_found = False
+
+    def warn(message: str):
+        nonlocal warnings_found
+        warnings_found = True
+        print(f"Warning: {message}")
```
joshuacwnewton, line 108: I imagine trying to deduplicate this function and dealing with module-level scope would be too messy to be worth it? :stuck_out_tongue_closed_eyes:

pyproject.toml
```diff
@@ -29,7 +29,6 @@ dependencies = [
   "nibabel",
   "opencv-python",
   "pandas",
-  "pandas-schema",
```
joshuacwnewton, line 32: Awesome! It seems like we're on the path to eliminating `pandas` too, eh? The [remaining usage](https://github.com/search?q=repo%3Aspine-generic%2Fspine-generic%20pandas&type=code) doesn't seem too bad!

joshuacwnewton (commented):
Lovely changes! Thank you for this. :hearts:

src/spinegeneric/cli/check_data_consistency.py
mguaypaq: In this case, it's because it's a double suffix, and `.with_suffix` only replaces (or removes) the last one. So this goes `*.nii.gz` -> `*.nii` -> `*.json`. Not the most elegant, but I think it's reasonably clear and concise.

src/spinegeneric/cli/check_data_consistency.py
mguaypaq: Yep! Using this kind of [type alias](https://docs.python.org/3/library/typing.html#type-aliases) is even more lightweight than using a [`NewType`](https://docs.python.org/3/library/typing.html#newtype). It can serve as both shorthand and documentation of intent. (And since we're not using any type checkers, our only use of type hints is for documentation.)

src/spinegeneric/cli/check_data_consistency.py
mguaypaq: I wondered about this as well. But then I thought that the main purpose of this code is to serve as an example of custom validation that other projects might develop for themselves. As such, I think this less general, more ad-hoc kind of validation is a more useful template.

src/spinegeneric/cli/check_data_consistency.py
mguaypaq: Well, it wasn't part of the original spec in this project, but it's [part of BIDS](https://bids-specification.readthedocs.io/en/stable/glossary.html#objects.columns.sex). I figured it's ok to have a custom validator be more restrictive than the spec (by not allowing lowercase or full words), but we might as well be inclusive because a definite "other" is not the same thing as a missing data "n/a".

src/spinegeneric/cli/check_data_consistency.py
mguaypaq: Yep! This filtering of `n/a` is done elsewhere, because it seemed like a parsing concern (it always means "missing data"), rather then a validation concern, but honestly it could have been either way:

https://github.com/spine-generic/spine-generic/blob/c52579386a82e5bb4f69043645c49d2a2d781c94/src/spinegeneric/cli/check_data_consistency.py#L144-L145

Still, I wanted to explicitly distinguish between `O` and `n/a` in the user-facing error message.

src/spinegeneric/cli/check_data_consistency.py
joshuacwnewton: Gotcha! Makes sense. :)

src/spinegeneric/cli/check_data_consistency.py
joshuacwnewton: Lovely!!

joshuacwnewton (approved):
Thanks for addressing my questions! LGTM. :)

src/spinegeneric/cli/check_data_consistency.py
mguaypaq: I figure that the performance hit from defining the inner helper function each time is tiny compared to the rest of what `main` does, but yes this is not my favourite pattern.

I still think it's more light-weight and explicitly local than the two other approaches that came to my mind:
- Defining `warnings_found` as a module-global variable, but philosophically the right scope for it is "every execution of `main`" rather than "the first time this module is imported".
- Defining it as a class with a single attribute and a single function, and instantiating it in `main`, which is too much ceremony for my taste.

src/spinegeneric/cli/check_data_consistency.py
mguaypaq: Thanks! Done in https://github.com/spine-generic/spine-generic/pull/287/commits/34cf0daead6c71d1f3b16867da4564171aed0519.

PR #288, opened 2025-05-27
Get rid of `str.strip`, `str.rstrip` bugs

mguaypaq:
For the `prog=` cases, it's unnecessary because the entry points from `pip install` are all `sg_*` names without a `.py` suffix already.

For the file name extension manipulation, the intended function is `str.removesuffix`, which removes the given string from the end if present.

(I didn't dare touch more of the code, since there are essentially no tests, and I've never had to run the scripts, so I don't know what the assumptions or expected outcomes are.)


joshuacwnewton (approved):
Thank you for the minor fix! I agree that these should be safe to do. :)

